PR #2, opened 2018-11-09
Add Appveyor config

sake92:
I guess Appveyor will trigger on next PR when this is merged..?  
Tests pass locally, so it should be alright.

### DIFF
--- a/os/src/os/ProcessOps.scala
+++ b/os/src/os/ProcessOps.scala
@@ -218,13 +218,19 @@ case class proc(command: Shellable*) {
 
     builder.directory(Option(cwd).getOrElse(os.pwd).toIO)
 
-    val cmd = 
-      if(scala.util.Properties.isWin) Vector("cmd.exe", "/C") ++ command.flatMap(_.value)
-      else command.flatMap(_.value)
+    // for details see https://github.com/lihaoyi/os-lib/pull/2#issuecomment-437873821
+    val cmds = command.flatMap(_.value)
+    val cmd = cmds(0)
+    val preparedCmds = if(!scala.util.Properties.isWin)
+      cmds
+    else if(os.proc.WinInternalCmdCommands contains cmd.toUpperCase) 
+      Vector("CMD.EXE", "/C") ++ cmds
+    else
+      cmds
 
     lazy val proc: SubProcess = new SubProcess(
       builder
-        .command(cmd:_*)
+        .command(preparedCmds:_*)
         .redirectInput(stdin.redirectFrom)
         .redirectOutput(stdout.redirectTo)
         .redirectError(stderr.redirectTo)
@@ -242,3 +248,11 @@ case class proc(command: Shellable*) {
   }
 }
 
+object proc {
+  // https://blog.brainasoft.com/all-internal-commands-of-cmd/
+  val WinInternalCmdCommands = Set(
+    "ASSOC","CALL","CD","CLS","COLOR","COPY","DATE","DEL","DIR","ECHO","ERASE","EXIT",
+    "FOR","FTYPE","MD","MKLINK","MOVE","PATH","PAUSE","POPD","PROMPT","PUSHD",
+    "REN","RD","SET","START","TIME","TITLE","TYPE","VER","VOL"
+  )
+}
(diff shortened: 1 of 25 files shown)


os/src/os/ProcessOps.scala
```diff
@@ -218,9 +218,13 @@ case class proc(command: Shellable*) {
 
     builder.directory(Option(cwd).getOrElse(os.pwd).toIO)
 
+    val cmd = 
+      if(scala.util.Properties.isWin) Vector("cmd.exe", "/C") ++ command.flatMap(_.value)
```
lihaoyi: I don't really understand what's going on here. Why doesn't normal subprocess execution work on windows?

sake92: I don't understand it either... When running just "dir" command, it just says "Command not found". With this approach I don't get that problem. Everyone recommends to run CMD commands that way..  

I digged this in [Apache Commons Exec](https://github.com/apache/commons-exec/blob/1.3/src/main/java/org/apache/commons/exec/launcher/WinNTCommandLauncher.java#L60). Seems like it is related to "current folder", in code they say:

>         // Use cmd.exe to change to the specified directory before running
>         // the command
>         final CommandLine newCmd = new CommandLine("cmd");
>         newCmd.addArgument("/c");
>         newCmd.addArguments(cmd.toStrings());

sake92:
First of all, what a mess... :smile:  
But I think I finally figured it out. These contribute to the problem:

1.  There is a [list of commands](https://blog.brainasoft.com/all-internal-commands-of-cmd/) that are internal to Windows' `CMD.EXE` command shell.  
    Those are **not physically present** on the disk, they are **not executables**!  
        (you can call them only in context of `CMD.EXE`).

2.  GitBash/Cygwin adds some Unix **executables** on the PATH...  
    They are found in: "C:\Program Files\Git\usr\bin".  
    Some of those are `dir.exe`, `sh.exe`, `perl.exe` etc.
    Btw, I didn't know there is a `dir` command on Unix... (same as `ls -C -b`)

Since I do have GitBash installed, I was confused with this:  
Command `cmd /c dir` called Windows' `DIR` command, but `dir` called Cygwin's `DIR.EXE`.  
Windows version returned verbose list, and Unix the regular, tabular data.

So, the solution could be to track calls to there Windows internal commands, something like this:

```scala
val cmd = "dir" // user calls this command
val WinInternalCmdCommands = Set("dir", "type", ...)
if(WinInternalCmdCommands contains cmd.toLowerCase) { // if internal
  Vector("cmd", "/c", cmd) // we prefix it with "cmd /c" call
} else {
  cmd
}
```

I'm only worried why no one does it this way...  
Anyone?

philwalk:
If I understand your solution, it will cause everyone working in Windows environment to get a CMD.EXE shell.    I suspect that many (maybe most?) developers would prefer this library to be more os-agnostic, so that the build.sc works the same way whether in linux, os-x or cygwin.     So here's a suggested alternative approach.

A more flexible test permits Windows uses to decide whether they want to spawn via CMD.EXE, bash.exe, or some other shell.   Rather than the following test:

`if(scala.util.Properties.isWin) `

let uses control behavior with a System property.   Here's how a user might specify to use bash via JAVA_OPTS:

`JAVA_OPTS=-Dos.lib.shell="/bin/bash -c"`

Here's the alternate code for defining the cmd:

```
val customShell = Option(sys.props("os.lib.shell"))
val cmd = (scala.util.Properties.isWin,customShell) match {
  case (_,Some(shell)) => shell.split(" ").toVector ++ command
  case (true,None) => Vector("cmd.exe", "/C") ++ command
  case (false,None) => Vector("/bin/sh","-c") ++ command
}
```

sake92:
@philwalk you'd get a CMD shell only when invoking one of *internal commands* (`dir`, `cd` etc.).  
I think that's what most users would expect when calling `proc("dir").call()`.  

When you invoke a command which is not one of those (e.g. `xcopy.exe` which is on `PATH`), you would do `proc("xcopy", "C:\\utils", "C:\\utils2").call()`. That would spawn a "normal" process, without `cmd /c`  stuff.

Relevant code snippet:
https://github.com/lihaoyi/os-lib/blob/0ad63bff5c27598c6236114f6b6124f455ab6109/os/src/os/ProcessOps.scala#L221-L229

lihaoyi:
@sake92 I have given you commit access to the repo; feel free to use it

PR #26, opened 2019-08-06
ProcessOutput.CallbackOutput needs to extend ProcessOutput.

lallea:


lallea:
The tests for scala 2.12 pass locally. The CI failures seem to be unrelated to this patch. I don't understand the build structure, so I don't know how to fix them.

PR #40, opened 2020-05-26
Buffer output streams in os.write for better performance

ajrnz:
This PR requires https://github.com/lihaoyi/geny/pull/11 to get the performance gains

`os.write` has poor performance when writing long iterable sequences eg

```scala
val lines = Iterator.range(0,1000000).map{x => x.toString + "\n"}.toSeq
os.write.over(os.pwd/"big.txt", lines)
```

which is about 10-30x slower than

```scala
os.write.over(os.pwd/"big.txt", lines.mkString("\n"))
```
This is due to repeated flushing of the main `OutputStream` in geny `Writable` but also the lack of buffering in `os.write`.

Here I buffer the output stream and also flush it at the end to ensure a full write. I elected not to close the output stream because that would also close the downstream classes which may not be desirable.

### DIFF
--- a/os/src/ReadWriteOps.scala
+++ b/os/src/ReadWriteOps.scala
@@ -10,6 +10,7 @@ import geny.Generator
 
 import scala.io.Codec
 import StandardOpenOption.{CREATE, WRITE}
+import java.io.BufferedOutputStream
 
 /**
   * Write some data to a file. This can be a String, an Array[Byte], or a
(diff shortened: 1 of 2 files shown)


lihaoyi:
Merged part of this as part of https://github.com/lihaoyi/os-lib/commit/736d651637f9740c3ba7ceceeff6e93d411cc1c7, to buffer in places where we can control the closing of the stream. OTOH I'm leaving `write.outputStream` as unbuffered by default, since you can always manually wrap it yourself if necessary

PR #50, opened 2020-12-17
Push all bytes when reading from a file channel

jodersky:
The FileChannel API makes no gurantees that a single call to "transferTo" will actually transfer all bytes from the file. In particular, if the file is large, it is likely that the destination channel will not accept all bytes at once.

### DIFF
--- a/os/src/Source.scala
+++ b/os/src/Source.scala
@@ -42,8 +42,13 @@ trait Source extends geny.Writable{
       os.flush()
     case Right(channel) =>
       (channel, out) match {
-        case (src: FileChannel, dest) => src.transferTo(0, Long.MaxValue, dest)
-        case (src, dest: FileChannel) => dest.transferFrom(dest, 0, Long.MaxValue)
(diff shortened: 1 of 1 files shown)


jodersky:
Note that the failure appears to be an issue of how mill handles dead symlinks in PathRefs. There's a fix on the way https://github.com/lihaoyi/mill/pull/1082. Alternatively, all tests pass with an older version of mill (0.8.0) which however is incompatible with Scala 3.0.0-M2.

jodersky:
Tests now pass with the updated mill version

lefou:
Can it be related to https://github.com/lihaoyi/mill/issues/528 ? We concat some shell script with a jar content. This seem to fail for large (maybe > 100MB) jar files. Here's the code:

```scala
os.write(output, prependShellScript + lineSep)
os.write.append(output, os.read.inputStream(tmp))
```

jodersky:
That's definitely possible. The circumstance that caused be find and fix this bug were in the context of serving a large ~300MB file from cask.

PR #51, opened 2020-12-26
Upgrade Scala versions (includes 3.0.0-M3)

jodersky:


.mill-version
```diff
@@ -0,0 +1 @@
+0.9.4-29-9d1d6f
```
lolgab, line 1: Can you update the version in the `mill` script instead ? Just for consistency with the other repos.

build.sc
```diff
@@ -92,8 +96,8 @@ trait OsLibModule extends CrossScalaModule with PublishModule{
 
 trait OsLibTestModule extends ScalaModule with TestModule{
   def ivyDeps = Agg(
-    ivy"com.lihaoyi::utest::0.7.5",
-    ivy"com.lihaoyi::sourcecode::0.2.1"
+    ivy"com.lihaoyi::utest::0.7.6",
```
lolgab, line 99:
```suggestion
    ivy"com.lihaoyi::utest::0.7.7",
```

build.sc
```diff
@@ -1,11 +1,16 @@
...
+
+val scalaNative = "0.4.0"
+
+val crossScalaVersions = Seq(scala212, scala213, scala3)
 val crossNativeVersions = Seq(
-  "2.11.12" -> "0.3.9",
-  "2.11.12" -> "0.4.0-M2"
+  scala212 -> scalaNative,
+  scala213 -> scalaNative
 )
```
lolgab, line 13: I moved to this convention here:
```scala
val dottyVersions = sys.props.get("dottyVersion").toList

val scalaVersions = "2.12.13" :: "2.13.4" :: "3.0.0-M3" :: dottyVersions
val scala2Versions = scalaVersions.filter(_.startsWith("2."))

val scalaJSVersions = for {
  scalaV <- scala2Versions
  scalaJSV <- Seq("0.6.33", "1.4.0")
} yield (scalaV, scalaJSV)

val scalaNativeVersions = for {
  scalaV <- scala2Versions
  scalaNativeV <- Seq("0.4.0")
} yield (scalaV, scalaNativeV)
```
Would you mind updating here too?

lolgab (changes requested):
Left a view comments, if you don't mind.

.mill-version
jodersky: Sure, but then we rely on people always using the local mill wrapper, and it may lead to inconsistencies if people use a system-installed mill wrapper. I don't really mind either way.

lolgab:
@jodersky can you also set the jdk version to 8 in travis so we're sure that we don't hit the same problem we had before with Scala 3.0.0-M3 not binary compatible ?
You can copy paste this one: https://github.com/lihaoyi/fansi/blob/master/.travis.yml

.mill-version
lihaoyi: The normal `mill` executable actually scrapes the local `./mill` file for the version, so it should behave the same way either way

jodersky:
Thanks for the comments. Let me know what you think now

.travis.yml
```diff
@@ -1,6 +1,9 @@
 os: linux
 language: scala
 
+jdk:
+  - openjdk8
+
 script:
 - curl https://raw.githubusercontent.com/scala-native/scala-native/master/scripts/travis_setup.sh | bash -x
```
lolgab, line 8: This is not needed anymore. Set 
```
dist: bionic
```
too.

lolgab (changes requested):
Just a tiny little thing :-) 
Thank you for considering my review :-)

.travis.yml
jodersky: ah, good to know. I'll just copy-paste the one you mentioned verbatim

jodersky: Btw, why is `__.__.test` needed, instead of `__.test`? Isn't a double __ redundant?

jodersky:
Would it possible to publish a release with these changes?

.travis.yml
lolgab: Yes, you are right.

lolgab:
@jodersky 0.7.2 has been published from this branch and it's on maven central. Can you bump the `publishVersion` so the PR is ready to merge?

jodersky:
Will, do. Thanks for the quick response!

lolgab (approved):

jodersky:
I don't have the necessary permissions to merge this, but please feel free to merge it yourself

lolgab:
@jodersky I don't either.
Let's wait for @lihaoyi to merge it :)

PR #53, opened 2021-01-07
fix: InvalidPathException when parsing URI#getFile String on Windows

Iltotore:
This is a fix originally made for Mill ([mentioned issue](https://github.com/lihaoyi/mill/pull/1092)). It fixes InvalidPathException when a String produced using URI#getFile is passed in the `apply` method.


sake92:
@Iltotore why did you delete mill executable? :D  
Also, your regex is a bit weird, `A-z`?  
Also 2, you could `t.drop(1)` which is a bit more understandable..

Iltotore:
> @Iltotore why did you delete mill executable? :D
> Also, your regex is a bit weird, `A-z`?
> Also 2, you could `t.drop(1)` which is a bit more understandable..

- I delete the exécutable because I was using a custom one to work with BSP on Windows. But I though I didn't push it. It's a mistake and I'll correct it after night.
- What do you find "weird"
- Okay. Let me do that after night.

sake92:
> * What do you find "weird"

Weird is `A-z`. I never saw it used elsewhere.  
Specifically this: https://stackoverflow.com/a/4923397/4496364  
So it would fail for `"/[:/abc"` for example...

Also 3, if this is windows-specific, shouldn't we use `scala.util.Properties.isWin` condition?  
Maybe just `if(scala.util.Properties.isWin && t.startsWith("/")) t.drop(1)` ..

Iltotore:
> Maybe just `if(scala.util.Properties.isWin && t.startsWith("/")) t.drop(1)` ..

I'm not sure the first / is totally incompatible with Windows' path system. It's just a specific pattern: like `/C:/` so i would suggest `if(scala.util.Properties.isWin && t.matches("\\/[A-z]:\\/.*")) t.drop(1)`

> So it would fail for "/[:/abc" for example...
I'm not sure it's a valid path. I'm maybe wrong.

lefou:
As commented on the original issue (https://github.com/lihaoyi/mill/pull/1092), a test would be nice.

Iltotore:
> As commented on the original issue ([lihaoyi/mill#1092](https://github.com/lihaoyi/mill/pull/1092)), a test would be nice.

Once I'm available, I'll do it. Surely this week.

lefou:
CI failed because the re-added mill executable isn't executable anymore. Probably a `chmod +x mill` will do the trick.

Iltotore:
> CI failed because the re-added mill executable isn't executable anymore. Probably a `chmod +x mill` will do the trick.

Ok. It's 22:00 here. I'll do that tomorrow

Iltotore:
It seems I don't have access to build logs if anyone can tell me what gone wrong. The mill is executable has +x permissions.

lefou:
@Iltotore I'm overhauling the CI setup at the moment, after that we can rebase as see what's really failing.

lefou:
@Iltotore I reworked it, but we still have some constantly failing tests. But Windows CI succeeds now, so rebasing or merging would be good.

os/test/src/PathTests.scala
```diff
@@ -30,6 +30,9 @@ object PathTests extends TestSuite{
           assert(root/"omg" == Path("/omg"))
           assert(rel/"omg" == RelPath("omg"))
           assert(sub/"omg" == SubPath("omg"))
+        } else {
+          //Windows path consistency
+          assert(Path((root/"omg").toIO.toURL.getFile) == root/"omg")
```
lefou, line 35: Shouldn't this test run on all platforms, not just Windows?

Iltotore: Originally it was a Windows-specific bug fix but it should also suceed on unix system so I can move it after school

os/src/Path.scala
```diff
@@ -481,7 +481,8 @@ sealed trait PathConvertible[T]{
 
 object PathConvertible{
   implicit object StringConvertible extends PathConvertible[String]{
-    def apply(t: String) = java.nio.file.Paths.get(t)
+    def apply(t: String) =
+      java.nio.file.Paths.get(if(scala.util.Properties.isWin && t.matches("\\/[A-z]:\\/.*")) t.drop(1) else t)
```
lefou, line 485:
```suggestion
      java.nio.file.Paths.get(if(scala.util.Properties.isWin && t.matches("\\/[A-Za-z]:\\/.*")) t.drop(1) else t)
```

As @sake92  commented earlier, to avoid matching some non-alphabetic character.

lefou (commented):
We are almost there. 2 things, though.

First I'd like you to either rebase your PR on latest master branch or merge the latest master into your PR, as you prefer. That way, we see the effect of latest changes to CI in this PR.

Second, as commented below, please fix the regex.

lefou:
This is the test failure on windows: https://github.com/lihaoyi/os-lib/pull/53/checks?check_run_id=1829465255#step:6:164

You probably need to reformat your patch to not exceed the max line length of 100.

os/src/Path.scala
```diff
@@ -481,7 +481,8 @@ sealed trait PathConvertible[T]{
 
 object PathConvertible{
   implicit object StringConvertible extends PathConvertible[String]{
-    def apply(t: String) = java.nio.file.Paths.get(t)
+    def apply(t: String) =
+      java.nio.file.Paths.get(if(scala.util.Properties.isWin && t.matches("\\/[A-Za-z]:\\/.*")) t.drop(1) else t)
   }
```
lefou, line 486:
```suggestion
      java.nio.file.Paths.get(
        if(scala.util.Properties.isWin && t.matches("\\/[A-Za-z]:\\/.*")) t.drop(1)
        else t
      )
  }
```

Apparently, the build fails because of the too long line, which exceeded the line limit.

Iltotore:
Thank you for the commit. I was a bit busy due to exams so thank you for committing it.

lefou:
In mill we now found (https://github.com/com-lihaoyi/mill/pull/1285) the culprit (which is cal to `(URL).getFile`) and a proper workaround without any string mangeling involved. We should probably adapt this PR to the same approach.

sake92:
@lefou or maybe close this PR because it's not a valid Path.. 😄   
If `java.nio.file.Paths` says so, makes no sense to have these special cases..

lefou:
> @lefou or maybe close this PR because it's not a valid Path.. smile
> If `java.nio.file.Paths` says so, makes no sense to have these special cases..

@sake92 You're absolute right. That was my second though and the reason why I merge the fix in mill instead, where we control how to get the proper path from the URL. I'm for closing this, too, as we fixed the original issue already.

lefou:
@Iltotore Thanks for your efforts to fix an issue in mill.

PR #64, opened 2021-03-22
Fix build failures

utgheith:
Fixing 3 issues:

(1) SpawningSubProcessTests::"spawn callback" has a race condition that causes occasional failures.
      It should use `SubProcess#join` instead of `SubProcess#waitFor` in order to guarantee that all stdout and stderr from the        subprocess is handled.

(2) Scala3 compilation fails when passing the "-release" flag to the compiler

(3) There is no macOS-11.0 support in GitHub Actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -5,7 +5,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


lefou (approved):
Looks good to me. Thank you!

PR #70, opened 2021-04-01
[WIP -- Please don't merge] NotifyWatcher

utgheith:
An `inotify` based watcher. Still rough around the edges and not fully stress-tested. Hoping for some early design/code review.

Why?

- Doesn't reorder/drop events (assuming no overflow)
- gives us much finer control/information about events


lihaoyi:
@utgheith I haven't gone through this in detail, but the overall approach looks reasonable

lefou:
What's the status of this PR? (And who is the main-maintainer of os-lib?) Is anybody OK with me merging it? I'd rather like to, as it would also solve the currently very flaky CI, which blocks PR #53 from getting green, which itself fixes a BSP bug in mill.

lihaoyi:
@lefou I suppose I'm still the main maintainer, but if you are willing to give it a review and @utgheith is happy with the PR go ahead and merge it

utgheith:
How about if we started by making it opt-in and enable it by default once it has been tested extensively?

Something like:

```os.watch2(...)```

or

```os.watch.enable_linux_inotify = true```

Rationale:

- Gives users (me included) an opportunity to test it in a real-world environment
- Allows us to evolve the API to allow for better error handling (e.g. https://github.com/com-lihaoyi/os-lib/pull/71)

**Comment by utgheith** [comment]
Added `os.watch.use_linux_inotify` to ease the transition. It defaults to `false` but the `os.watch` tests set it to `true`

File: os/watch/src/package.scala [hunk 1]
```diff
@@ -45,4 +52,6 @@ package object watch{
     thread.start()
     watcher
   }
+
+  var use_linux_inotify: Boolean = false
```
lefou, line 56: I don't like the idea of switching this globally. I'd prefer a parameter which enables "native" support. The scaladoc should explain what "native" means for each platform. Also, making this binary compatible would be good idea, e.g. keeping the old `def watch` signature with sensible default to current behavior.

There is one issue though, as for Mac OS there seems to be be currently no choice but only the native implementation. I can't tell if it's because the native version just works better or because the generic version doesn't work at all.

build.sc
```diff
@@ -116,7 +121,7 @@ trait OsLibTestModule extends ScalaModule with TestModule{
     millSourcePath / s"src-$platformSegment"
   )
 
-  def testFrameworks = Seq("utest.runner.Framework")
+  def testFrameworks = Seq("utest.runner.Framework", "org.scalacheck.ScalaCheckFramework")
```
lefou, line 124: We just deprecated muitiple test frameworks per test module in mill. Either call ScalaCheck API from utests or create a seconds test module using the other test framework.

os/watch/src/main.scala
```diff
@@ -0,0 +1,8 @@
+
```
lefou, line 1: What is this files. Shouldn't it reside in the test module?

lefou (commented):
I haven't reviewed the actual inotify implementation, just the integeration for now.

os/src/Internals.scala
```diff
@@ -22,4 +22,8 @@ object Internals{
     src,
     dest.write(_, 0, _)
   )
+
+  def linux() = {
+    System.getProperty("os.name") == "Linux"
+  }
```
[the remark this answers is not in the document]
utgheith: The problem is that it didn't exist in 2.11. It is used in exactly one place. I could easily move the test there and remove this method.

os/watch/src/package.scala
utgheith: I did it this way in order to stress the fact that it's an internal implementation detail not intended to survive for a long time and should be neither documented not show up on the public API.

A reasonable alternative is to add an extra argument to `os.watch`. Naming it something like `preferNative` would suggest that it's just a preference. It will always be picked on MacOS because it's the only choice.

build.sc
utgheith: I will do some digging and figure out how to do that

os/watch/src/main.scala
utgheith: Ops! Will remove it.

os/src/Internals.scala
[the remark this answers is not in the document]
utgheith: https://github.com/com-lihaoyi/os-lib/pull/70/commits/a87c11fc917bee5304774c287bf496fa66ba8fc1

os/watch/src/package.scala
utgheith: https://github.com/com-lihaoyi/os-lib/pull/70/commits/3760863210ff8d9af4466d4d8b2c70e1da2eb6e2

os/watch/src/main.scala
utgheith: https://github.com/com-lihaoyi/os-lib/pull/70/commits/c2c1c63aba37ef61159a1037bf31222488049458

build.sc
utgheith: https://github.com/com-lihaoyi/os-lib/pull/70/commits/99ae578fb2730b0c81afa1ed61dc3d623994db92

os/watch/src/package.scala
lihaoyi: @lefou the generic OSX version uses polling on every watched file, which is as good as "doesn't work at all" on non-trivial folder hierarchies

os/src/Internals.scala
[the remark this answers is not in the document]
lefou: You're right, I didn't though about older Scala versions. Personally, I'd first move the "Internals" object into an "internal" package to make sure nobody outside uses it. Then we can decide whether we want to provide the check as an `def isLinux: Boolean` or if we want to provide different versions per Scala-version and either delegate or self-implement.

os/watch/src/package.scala
lefou: I like the idea with the `WatchConfig`, as it avoids future changes to `watch` API for such kind of tweaks. I don't like the `deprecated` annotation without any hint though. Also, we should give the user the fine control she wants, and make a flag per platform, when appropriate. And after @lihaoyi s comment I think it will be end up with only one flag right now: `useLinuxInotify`.

lefou: I think when we make the new `NotifyWatcher` optional and disabled by default, this is good enough to merge as-is. I doubt, we will find a reviewer familiar enough with Linux inotify system to review this timely.

build.sc
```diff
@@ -104,6 +105,10 @@ trait OsLibTestModule extends ScalaModule with TestModule{
   def ivyDeps = Agg(
     ivy"com.lihaoyi::utest::0.7.11",
     ivy"com.lihaoyi::sourcecode::0.2.8"
+    if (scalaVersion().startsWith("2.11"))
+       ivy"org.scalacheck::scalacheck::1.15.2"
+    else
+       ivy"org.scalacheck::scalacheck::1.15.3"
```
lefou, line 111: Missing comma

```suggestion
    ivy"com.lihaoyi::utest::0.7.11",
    ivy"com.lihaoyi::sourcecode::0.2.8",
    if (scalaVersion().startsWith("2.11"))
       ivy"org.scalacheck::scalacheck::1.15.2"
    else
       ivy"org.scalacheck::scalacheck::1.15.3"
```

build.sc
```diff
@@ -46,7 +46,8 @@ object os extends Module {
       def platformSegment = "jvm"
       def moduleDeps = super.moduleDeps :+ os.jvm()
       def ivyDeps = Agg(
-        ivy"net.java.dev.jna:jna:5.0.0"
+        ivy"net.java.dev.jna:jna:5.0.0",
+        ivy"com.lihaoyi::sourcecode::0.2.5",
```
lefou, line 50:
```suggestion
        ivy"com.lihaoyi::sourcecode::0.2.8",
```

build.sc
```diff
@@ -103,7 +104,11 @@ trait OsLibModule extends CrossScalaModule with PublishModule{
 trait OsLibTestModule extends ScalaModule with TestModule{
   def ivyDeps = Agg(
     ivy"com.lihaoyi::utest::0.7.11",
-    ivy"com.lihaoyi::sourcecode::0.2.8"
+    ivy"com.lihaoyi::sourcecode::0.2.8",
+    if (scalaVersion().startsWith("2.11"))
+       ivy"org.scalacheck::scalacheck::1.15.2"
+    else
+       ivy"org.scalacheck::scalacheck::1.15.3"
```
lefou, line 111:
```suggestion
       ivy"org.scalacheck::scalacheck::1.16.0"
```

Issue #78, opened 2021-05-20
File.getCanonicalFile "normalizes" more than Path.normalize on Windows


alexarchambault:
My Windows user name is `Alexandre`, which is more than 8 characters long. Creating temporary directories gives me paths which are in ["8.3" format](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#short-vs-long-names):
```text
C:\Users\Alexandre>amm -c "println(os.temp())"
Compiling C:\Users\Alexandre\(console)
C:\Users\ALEXAN~1\AppData\Local\Temp\2085325879310438089.tmp
```
(with `ALEXAN~1` rather than `Alexandre` in the path).

In this case, we have `p != os.Path(p.toIO.getCanonicalFile)`, which is surprising I think:
```text
C:\Users\Alexandre>amm -c "val p = os.temp(); println(p == os.Path(p.toNIO.normalize)); println(p == os.Path(p.toIO.getCanonicalFile))"
Compiling C:\Users\Alexandre\(console)
true
false
```

That originates from the different handling of the 8.3 / long paths, calling `java.io.File.getCanonicalFile` on those files rather than `java.nio.file.Path.normalize` returns different paths:
```text
C:\Users\Alexandre>amm -c "val p = os.temp(); println(\"NIO: \" + p.toNIO); println(\" IO: \" + p.toIO.getCanonicalFile)"
Compiling C:\Users\Alexandre\(console)
NIO: C:\Users\ALEXAN~1\AppData\Local\Temp\2933120367880475586.tmp
 IO: C:\Users\Alexandre\AppData\Local\Temp\2933120367880475586.tmp
```

It seems `File.getCanonicalFile` calls [`GetLongPathName`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getlongpathnamea) while `Path.normalize` doesn't.

I don't know if anyone has an opinion about this. I'd argue the paths returned by `getCanonicalFile` are more "normal", and should be used by default here. In particular, I've had issues with a downstream library calling `getCanonicalFile` at some point, which returned apparently different paths than os-lib.

I can send a PR to replace the few uses of `Path.normalize` by `File.getCanonicalFile` if everyone is fine with it.

lefou:
[...]

PR #79, opened 2021-05-23
Prefer File.getCanonicalFile over Path.normalize

alexarchambault:
Fixes https://github.com/com-lihaoyi/os-lib/issues/78.

### DIFF
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -1,6 +1,7 @@
 package os
 
 import collection.JavaConverters._
+import util.Properties
 
 trait PathChunk{
   def segments: Seq[String]
@@ -384,7 +385,20 @@ object Path {
       throw PathError.AbsolutePathOutsideRoot
     }
 
-    val normalized = f.normalize()
+    fromNIO(f)
+  }
+
+  private def fromNIO(p: java.nio.file.Path): Path = {
+    // On Windows, File.getCanonicalFile normalizes 8.3 paths as long paths,
+    // whereas Path.normalize doesn't. So we use the former here.
+    val shouldUseIO = Properties.isWin &&
+      p.getFileSystem.getClass.getName == "sun.nio.fs.WindowsFileSystem"
+    val normalized =
+      if (shouldUseIO) {
+        require(p.isAbsolute, s"$p is not an absolute path")
+        p.toFile.getCanonicalFile.toPath
+      }
+      else p.normalize()
     new Path(normalized)
   }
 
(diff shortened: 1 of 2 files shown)


alexarchambault:
This might be a problem if people use os-lib with non-standard file systems, like NIO allows… Sometimes, the `Path` from these file systems correspond to purely virtual things, and cannot be converted to a concrete `File`. Don't know if that's a use case that os-lib should handle…

lefou (commented):
Looks good to me. Having a test for it, would make me more comfortable.

alexarchambault:
> Having a test for it, would make me more comfortable.

Same for me, actually… I'll try to add one the next time I'll boot my Windows machine.

alexarchambault:
Ok, I added a test… and now I'm shared about merging this 😬

`File.getCanonicalFile` depends on what's on disk, so that two calls of `os.Path(…).toString` may not return the same value (if we pass it a short form, it will return a longer one, but only when the corresponding file exists), which could be surprising.

Also, it seems `File.getCanonicalFile` caches its values: if it's called with a short form before the file exists, it will later always return the short form, even if we create the underlying file later on. Which is equally surprising.

`Path.normalize` doesn't suffer those quirks, AFAIK.


And fwiw, a similar issue can be reproduced on macOS (with its "case-insensitive but case-preserving" file systems), using just the case:
```text
$ mkdir Foo && amm -c 'println((os.pwd / "foo").toIO.getCanonicalFile)'
/Users/alexandre/test/Foo

$ rmdir Foo && amm -c 'println((os.pwd / "foo").toIO.getCanonicalFile); os.makeDir(os.pwd / "Foo"); println((os.pwd / "foo").toIO.getCanonicalFile)'
/Users/alexandre/test/foo
/Users/alexandre/test/foo

$ ls
Foo
```

lefou:
> Ok, I added a test… and now I'm shared about merging this grimacing

Do you mean "scared"?

> `File.getCanonicalFile` depends on what's on disk, so that two calls of `os.Path(…).toString` may not return the same value (if we pass it a short form, it will return a longer one, but only when the corresponding file exists), which could be surprising.

Oh, as I commented on #78, I'd rather remove any dependency on disk state, and I was under the impression, you're going to always normalize to the long form. In that regard, I don't understand the test case, at least, it test something different.

> But a worse issue is that it seems that `File.getCanonicalFile` caches its values: if it's called with a short form before the file exists, it will later always return the short form, even if we create the underlying file later on.
> 
> `Path.normalize` doesn't suffer those quirks, AFAIK.

Better, we don't merge this.

alexarchambault:
It'd be better not to merge, yes.

(And it seems there's a need for a library providing true canonical paths, without those caching quirks…)

PR #81, opened 2021-06-14
Improve Internals.transfer0 by declaring r within the loop

Flowdalic:
By declaring r within the loop, we reduce its scope, can declare it
immutable, and shorten the function. Furthermore, r is not checked
twice (even though an optimizing compiler may eliminates the second
check).

### DIFF
--- a/os/src/Internals.scala
+++ b/os/src/Internals.scala
@@ -1,5 +1,7 @@
 package os
 
+import scala.util.control.Breaks._
+
 import java.io.{InputStream, OutputStream}
 import java.nio.file.Files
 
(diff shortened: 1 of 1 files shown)


lihaoyi:
Probably going to call this a wontfix; the code is fine as it is now, and introducing a try-catch flow as part of `breakable` doesn't seem like an improvement

Flowdalic:
I wasn't aware that `break` is syntactic sugar in Scala when I started this, and, coming from a systems background, the implementation just looked off. I agree that the proposed implementation isn't nice either, so it's a question of which one you like less. :)

PR #88, opened 2021-12-09
Add os.Path.of()

Flowdalic:


Flowdalic:
The same CI run succeeds in my fork: https://github.com/Flowdalic/os-lib/actions/runs/1558596640
Is this a flapping test?

Flowdalic:
Happy new year ping. 🙃

lefou (commented):
I like the change to `os.pwd`, as it reflects the fact that a process working dir can change (although this is rarely seen).

The `Path.of` looks just like some convenience to me. If feels a bit inconsistent (naming wise), as we already have some overloads of the apply method. Also, it's implementation is much copy and paste of functionality already present in other `def `s, e.g. `expandUser`.

Flowdalic:
> I like the change to `os.pwd`, as it reflects the fact that a process working dir can change (although this is rarely seen).

Glad you like it. I actually believe that this is a straight bug in os-lib. Especially if you use native libs, chances increase that your cwd will not be stable across the process's lifetime. So I am not sure if I would agree to this being rarely seen.

> The `Path.of` looks just like some convenience to me.

It absolutely is, but IMHO it's fills a whole in the API of os-lib. I was really surprised to that it took me a while to find out how to glue the different API parts together so that the behave like my suggested `Path.of()` implementation.


> If feels a bit inconsistent (naming wise), as we already have some overloads of the apply method.

Any naming suggestions would be appreciated. I was looking for a short method name. Note that the Java Runtime Library also provides [`Path.of()`](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/nio/file/Path.html#of(java.lang.String,java.lang.String...)), although with slightly different semantics. Hence I am not sure if this is an argument for or against re-using `of()`, here.

> Also, it's implementation is much copy and paste of functionality already present in other def s, e.g. expandUser.

It is, unfortunately I believe, re-using the code of `expandUser` would mean to factor out the re-usable code into a new method (or maybe my Scala Foo is to weak). In order to keep the PR's changes minimal, I decided to refrain from doing so, as I figured it could be done later anyways.

So how do we proceed with this?

lefou:
@Flowdalic You may split your PR, to get the "bug" fixed. This results in a breaking change though. The code duplication IMHO requires a change. Decision about the name and the support of the special tilde ("~") should be made by a maintainer though.

Flowdalic:
Created https://github.com/com-lihaoyi/os-lib/pull/94 and awaiting @lihaoyi's feedback.

lihaoyi:
`Path.of` seems useful, but the name is bad. It is basically identical to `Path.apply(_, os.pwd)`, except it expands a leading `~` into the user's home directory. Maybe we could add another flag to `Path.apply` to do this? `Path.apply(_, os.pwd, expandHome = true)`?

lefou:
I like the extra `expandHome: Boolean` parameter. Clear and consistent.

lihaoyi:
Some precedence is that Python calls it [os.path.expanduser](https://docs.python.org/3/library/os.path.html#os.path.expanduser), although they support `~user` in addition to `~`, and it's a transformation rather than a constructor

lefou:
I wonder if `def expandUser` isn't already what we try to add here, except that relative paths need an explicit base, which is IMHO ok. In other words, proposed `Path.of()` is equivalent to `Path.expandUser(_, os.pwd)`. This was not obvious to me before.

lihaoyi:
Hah I had totally forgotten about the `def expandUser` function we already have. Yes, it looks basically identical to what is being proposed

Flowdalic:
> Hah I had totally forgotten about the `def expandUser` function we already have. Yes, it looks basically identical to what is being proposed

Except when it is not. :) Note that my main motivation of `Path.of()` is, as stated in the commit message of the related commit: "In particular, there is no function that takes either an absolute or relative path input and constructs an absolute os.Path from it."

> In other words, proposed `Path.of()` is equivalent to `Path.expandUser(_, os.pwd)`

Hence, if I am not mistaken, `Path.of()` is *not* equivalent to `Path.expandUser(_, os.pwd)`: If the input is `/bar`, and the current working directory is `/foo`, then `Path.of()` would return `/bar`, while `Path.expandUser(_, os.pwd)` would  return `/foo/bar`.

I write a lot of code where users configure a directory either relative to the current working directory, relative to their home directory, or as an absolute path. And os-lib lacks a method that simply takes the user inputted String in any of those three forms and returns an absolute Path object.

I hope that helps to understand the motivation for this change.

lefou:
> Hence, if I am not mistaken, `Path.of()` is _not_ equivalent to `Path.expandUser(_, os.pwd)`: If the input is `/bar`, and the current working directory is `/foo`, then `Path.of()` would return `/bar`, while `Path.expandUser(_, os.pwd)` would return `/foo/bar`.

This is not correct. The path given with `base` is only prepended if the given path is a `RelPath`. If you provide an absolute path, and `/bar` is absolute, the base path will not be prepended.

Flowdalic:
Ahh, thanks for pointing this out. Seems like this should turn into a change to improve the documentation. 😀

lihaoyi:
I guess we can close this then?

Flowdalic:
> I guess we can close this then?

Yes, thanks for the help. :)

PR #89, opened 2021-12-09
Make os.remove behave more like Files.deleteIfExists, add checkExists flag to fall back to old behavior

lihaoyi:
Follow up from the discussion in https://github.com/com-lihaoyi/os-lib/pull/86. As discussed, we hope that it would be more intuitive to avoid failing loudly if a file we want to delete already doesn't exist. If the user cares, they can always check the returned `Boolean`, or pass `checkExists = true` if they want the exception

CC @iulian-db

Review by @lefou @sake92

### DIFF
--- a/os/src/FileOps.scala
+++ b/os/src/FileOps.scala
@@ -280,8 +280,16 @@ object copy {
  * any files or folders in the target path, or
  * does nothing if there aren't any
  */
-object remove extends Function1[Path, Unit]{
-  def apply(target: Path): Unit = Files.delete(target.wrapped)
+object remove extends Function1[Path, Boolean]{
+  def apply(target: Path): Boolean = apply(target, false)
+  def apply(target: Path, checkExists: Boolean = false): Boolean = {
+    if (checkExists) {
+      Files.delete(target.wrapped)
+      true
+    }else{
+      Files.deleteIfExists(target.wrapped)
+    }
+  }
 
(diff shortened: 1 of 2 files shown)


os/src/FileOps.scala
```diff
@@ -280,8 +280,16 @@ object copy {
  * any files or folders in the target path, or
  * does nothing if there aren't any
  */
-object remove extends Function1[Path, Unit]{
-  def apply(target: Path): Unit = Files.delete(target.wrapped)
+object remove extends Function1[Path, Boolean]{
+  def apply(target: Path): Boolean = apply(target, false)
+  def apply(target: Path, checkExists: Boolean = false): Boolean = {
```
sake92, line 285: Do we need 2 apply-es if we have a default value for `checkExists: Boolean = false`? 😅

sake92 (approved):

os/src/FileOps.scala
lefou: This one is needed for binary compatibility, and probably also for easier applicability, e.g. in `files.map(os.remove)`.

lefou (approved):
LGTM. We should bump the minor version when releasing.

os/src/FileOps.scala
sake92: Ah, ok, thx.

Issue #83, opened 2021-09-09
More info in SubprocessException


DavidPerezIngeniero:
In case we launch lots of commands, it would be useful that SubprocessException had information about the command launched with its arguments.
It would be easier to diagnose what has failed.

DavidPerezIngeniero:
Here is the pull request:
https://github.com/com-lihaoyi/os-lib/pull/93

lefou:
Implemented by #100

PR #93, opened 2021-12-28
Informationm about the command that has crashed

DavidPerezIngeniero:
This small PR provides more information about which command has crashed.
When you laumch many commands in complex apps, you don't know what has failed by looking at the exception.
https://github.com/com-lihaoyi/os-lib/issues/83

### DIFF
--- a/os/src-jvm/ProcessOps.scala
+++ b/os/src-jvm/ProcessOps.scala
@@ -82,7 +82,7 @@ case class proc(command: Shellable*) {
     val chunksArr = chunks.iterator.asScala.toArray
     val res = CommandResult(sub.exitCode(), chunksArr)
     if (res.exitCode == 0 || !check) res
-    else throw SubprocessException(res)
+    else throw SubprocessException(res, command.flatMap(_.value))
   }
 
   /**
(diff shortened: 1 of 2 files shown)


lefou (commented):
This looks good to me. What I wonder is whether we should think about information that is not suspected to appear in the message, like passwords. We could just provide an explicit Shellable instance (`ProtectedShellable`) which only prints `"<protected>"` instead of it's content. It's just loud thinking right now. WDYT @lihaoyi ?

lihaoyi:
I think this looks reasonable. The concerns about passwords etc. are real though.

While a developer running code locally would already have access to anything the stack trace could show, lots of teams ship error logs to centralized aggregators (e.g. sentry.io or ELK) where you may be fine with a developer seeing stack traces but less fine with them seeing production credentials.

Similarly, many executables you may be spawning as subprocesses are outside your control, and thus you may be forced to pass in sensitive things via main arguments regardless of best practices. The best practices of not passing sensitive information via CLI parameters only really applies to shells which keep history, so there isn't any best practice not to pass in sensitive information via main args to subprocesses.

Could a compromise be we keep the metadata as part of the exception object, but not render it in the error message? That way anyone who wants can try-catch-pattern-match on the exception type and print it out, but it won't get scooped up accidentally by people who are aggregating their stack traces

PR #94, opened 2022-01-05
Fix os.pwd: evalute every time and use portable approach

Flowdalic:
The current working directory of a process may change over time, hence
os.pwd must not be a value that is initialized once, but instead a
callable function.

Furthermore, I think the more portable approach to get the current
working directory is

Paths.get("").toAbsolutePath()

as "." appears to be Unix specific.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -107,7 +107,8 @@ trait OsLibModule extends CrossScalaModule with PublishModule{
 trait OsLibTestModule extends ScalaModule with TestModule{
   def ivyDeps = Agg(
     ivy"com.lihaoyi::utest::0.7.10",
-    ivy"com.lihaoyi::sourcecode::0.2.7"
+    ivy"com.lihaoyi::sourcecode::0.2.7",
+    ivy"com.github.jnr:jnr-posix:3.1.14",
   )
 
   def platformSegment: String
--- a/os/src-jvm/package.scala
+++ b/os/src-jvm/package.scala
@@ -18,10 +18,16 @@ package object os{
    */
   val home: Path = Path(System.getProperty("user.home"))
 
+  private val initialWorkingDirectory: Path = os.Path(java.nio.file.Paths.get(".").toAbsolutePath)
+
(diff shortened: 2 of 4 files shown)


lihaoyi:
Can you provide a unit test that demonstrates the pwd changing over time? I have never seen that happen

Flowdalic:
> Can you provide a unit test that demonstrates the pwd changing over time? I have never seen that happen

Turns out that there is a plot twist. Java does not have a notion of a *current* working directory. The commonly referred idiom of  `Paths.get("").toAbsolutePath()` appears to only ever return what is probably best described as "initial working directory". However POSIX processes have a notion of a *current* working directory, and just like environment variables, its value can change over a processe's lifetime. If os-lib wants to have proper handling of the current working directory and especially the transformation of relative paths to absolute ones, then the currently implementation of os.pwd must be able to retrieve the actual current working directory.

I've updated the PR with my suggestion on how it could be done. Feedback is appreciated. :)

lihaoyi:
It sounds like sticking to the JVM standard is the right thing to do here. OS-Lib is meant to be a wrapper around the Java APIs rather than the POSIX standards, and of course runs on places like Windows where there is only a JVM and no POSIX compatibility. If someone wants a POSIX CWD they can always implement this in their project, as just another method returning an `os.Path`, but it doesn't feel like it should be part of OS-Lib itself

Flowdalic:
> It sounds like sticking to the JVM standard is the right thing to do here.

Wouldn't that, in last consequence, mean that `os.pwd` should be removed? After all, there is no such thing as working directory in Java. I think this should be at least documented.

lihaoyi:
Probably not worth removing `os.pwd`. It might not be *the* POSIX working directory, but it is what most JVM programmers would consider a working directory.

I think we can close this then. Was definitely a good discussion, and even if we didn't end up merging it we explored the design space more thoroughly than we had before, and at least I learned something from this :)

Flowdalic:
Dito. :)

PR #98, opened 2022-01-31
Bump some patch versions

jodersky:
This also fixes the issue with `docJar` under Scala Native (3.1.0, 4.3), which is currently preventing publishing.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -11,8 +11,8 @@ import de.tobiasroeser.mill.vcs.version.VcsVersion
 val dottyVersions = sys.props.get("dottyVersion").toList
 
 val scala2AndDottyVersions = "2.12.13" :: "2.13.4" :: "2.11.12" :: dottyVersions
-val scala30 = "3.0.0"
-val scala31 = "3.1.0"
+val scala30 = "3.0.2"
+val scala31 = "3.1.1"
 
 val scalaNativeVersions = for {
   scalaV <- scala31 :: scala2AndDottyVersions


lolgab (approved):
I removed the `0.8.1` tag so we can push it again since it wasn't released.

jodersky:
are the `macos-latest` CI tests important? They seem to have been failing for other PRs too

lolgab:
I have no idea. Probably @lefou knows more.

jodersky:
it looks like some of the tests are flaky

lefou:
> are the `macos-latest` CI tests important? They seem to have been failing for other PRs too

Yeah, these flaky test are an issue, but we should keep them. Maybe these can be improved or if known to be flaky, we can add some retry logic. I think we also have MacOSX specific code (or PRs), and we want os-lib to be used on MacOSX.

lefou:
If these flaky tests are blocking publishing, as a last resort we could let the publish job only depend on the linux tests.

jodersky:
Agreed that we should keep them. Could we change the CI logic that tests are only run on pull requests, and unconditionally publish when a tag is pushed? I think that we could reasonably assume that a pushed tag is always intended for release, and skip the testsuite altogether on such an event.

lolgab:
> I think that we could reasonably assume that a pushed tag is always intended for release, and skip the testsuite altogether on such an event.

Yes, but people make mistakes, let's have the other tests as a safeguard.

jodersky:
IMO it's ok to run tests on a tag push, but the publishing step should always occur. The thing is that once a tag is public, it's difficult to retract it. I know we kinda did it with this one :), and I hope no one relied on that tag from this weekend.

jodersky:
FYI, I had started a manual release process before I saw the convo here. 0.8.1 should be available on maven central soon. I published from a clean checkout of 0.8.1, commit sha fd793f9072ec0e9821907500eb093692bf4c974d

lolgab:
> and I hope no one relied on that tag from this weekend.

I don't think there are people using Scala libraries from sources..

lefou:
I don't like to remove tags too, but I also did it in Mill (within hours) because I don't had permission to release manually. 

How pressing is this release issue. I typically just re-run GHA to overcome flaky tests. Is this to working in os-lib? Is there a "most stable" platform or potentially a non-flaky sub-set of tests we can keep enabled as precondition for publishing from tags?

jodersky:
I tried running the tests around 4 times. The most flaky was the macos 1.8 test, but the others unfortunately also did not have a 100% success rate

lefou:
Looks like it's always the same test that fails (watch). Maybe we can somehow mark/tag these flaky tests and make their sucess optional when needed? I don't know how to do it in utest though.

lefou:
Or create a second test module in mill setup and move these tests over. Would be even better to just fix the tests, though.

lolgab:
Utest as a mechanism for flaky tests: https://github.com/com-lihaoyi/utest#suite-retries

lolgab:
I tried this: https://github.com/com-lihaoyi/os-lib/pull/99

jodersky:
ah yes, I should have said which tests are flaky. It's indeed always the "watch" tests, and the macos runner seems to be more prone to failure that the others, although they all fail on occasion. I wonder if this could be related to the fact that we're running these tests with github actions, which presumable runs them in a VM, on a machine which probably hosts many other VMs. When I run them on my laptop they never seem to fail.

lolgab:
@jodersky @lefou Created a PR to retry the WatchTests and the CI is green on that PR.
https://github.com/com-lihaoyi/os-lib/pull/99

PR #99, opened 2022-01-31
Retry flaky tests

lolgab:
This retries the flaky watch tests we have

### DIFF
--- a/os/watch/test/src/WatchTests.scala
+++ b/os/watch/test/src/WatchTests.scala
@@ -2,7 +2,10 @@ package test.os.watch
 
 import utest._
 object WatchTests extends TestSuite with TestSuite.Retries{
-  override val utestRetryCount = 10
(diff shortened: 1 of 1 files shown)


lefou (commented):
10 retries looks a bit too high and reduces my confidence.

Also, maybe we should make this conditional or configurable, e.g. by an env var `TEST_RETRY_COUNT` which we set in the build? We could check for `CI` or `GITHUB_ACTIONS` (https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables), so that we don't retry when testing locally (and possibly shadow malfunction).

lolgab:
I tried with 5 and it failed on Mac OS. Actually I can configure the retry to 10 only on mac and on Github Actions. What do you think @lefou ?

lefou:
I think Linux is also failing sometimes. If you retry only on GITHUB, I'm fine with it.

lefou (approved):
LGTM.

PR #100, opened 2022-02-21
Add the actual command to CommandResult

Flowdalic:
The messages of exceptions thrown by call() would previously just
mention the exit status of the invoked process, but not the actual
command line that was used to create the process. To help with
debugging purposes, the exception message now includes the full
command line and the os.proc instance provides access to it.

### DIFF
--- a/os/src-jvm/ProcessOps.scala
+++ b/os/src-jvm/ProcessOps.scala
@@ -22,6 +22,8 @@ import scala.annotation.tailrec
   *   want
   */
 case class proc(command: Shellable*) {
+  val commandChunks = command.flatMap(_.value)
+
   /**
     * Invokes the given subprocess like a function, passing in input and returning a
     * [[CommandResult]]. You can then call `result.exitCode` to see how it exited, or
@@ -80,7 +82,7 @@ case class proc(command: Shellable*) {
     sub.join(timeout)
 
     val chunksArr = chunks.iterator.asScala.toArray
-    val res = CommandResult(sub.exitCode(), chunksArr)
+    val res = CommandResult(commandChunks, sub.exitCode(), chunksArr)
     if (res.exitCode == 0 || !check) res
     else throw SubprocessException(res)
   }
@@ -117,7 +119,6 @@ case class proc(command: Shellable*) {
 
     builder.directory(Option(cwd).getOrElse(os.pwd).toIO)
 
-    val commandChunks = command.flatMap(_.value)
     val commandStr = commandChunks.mkString(" ")
     lazy val proc: SubProcess = new SubProcess(
       builder
--- a/os/src-jvm/ProcessOps.scala
+++ b/os/src-jvm/ProcessOps.scala
@@ -22,7 +22,7 @@ import scala.annotation.tailrec
   *   want
   */
 case class proc(command: Shellable*) {
-  val commandChunks = command.flatMap(_.value)
+  def commandChunks: Seq[String] = command.flatMap(_.value)
 
   /**
     * Invokes the given subprocess like a function, passing in input and returning a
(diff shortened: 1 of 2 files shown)


lefou:
This is related to

* #83 and
* #93 

and the discussion in PR #93 also applies to it. We should be careful to not expose secrets by print the whole command.

Flowdalic:
> We should be careful to not expose secrets by exposing the whole command.

I would love to life in a world where people know by now that secrets should not be part of the command, but I know that sadly, the reality is different. @lihaoyi suggested in https://github.com/com-lihaoyi/os-lib/pull/93#issuecomment-1006285002 a compromise to put the command in `CommandResult` but not show it in the exception message. I incorporated that, which the adjustment to display the first chunk of the command, which should be safe.

os/src-jvm/ProcessOps.scala
```diff
@@ -22,6 +22,8 @@ import scala.annotation.tailrec
   *   want
   */
 case class proc(command: Shellable*) {
+  lazy val commandChunks = command.flatMap(_.value)
```
lefou, line 25: Is it necessary to persist it? Can't it be a simple `def`?

lefou (commented):
Looks reasonable to me.

os/src-jvm/ProcessOps.scala
Flowdalic: I don't have a strong opinion on that, other than that it appears very sensible that os.proc() provides access to the actual command (for the same rationale that it is sensible that CommandResult not only contains the result but the actual command that caused the result).
I used lazy here because it appeared to be the idiomatic thing to do, even now that it's not a big calculation. And if the overhead of the synchronization imposed by lazily evaluation `commandChunks` becomes a bottleneck in an app, then the app has a fundamental design issue in that region. But that said, I am nowhere near to decide if `lazy` or `def` is more idiomatic in such cases and I am happy to hear the opinion of Scala veterans on this.

lefou: This is pure convenience, as `command` is already accessible, and easy to compute, so it's IMHO not worth the extra memory space, hence `def` is better.

Flowdalic: Fair point, switched to `def`.

Flowdalic:
Friendly reminder about this PR. :)

lefou:
Hm, looks like PR is too old to start CI...

Flowdalic:
> Hm, looks like PR is too old to start CI...

Anything I need to do?

os/src-jvm/ProcessOps.scala
```diff
@@ -117,7 +119,6 @@ case class proc(command: Shellable*) {
 
     builder.directory(Option(cwd).getOrElse(os.pwd).toIO)
 
-    val commandChunks = command.flatMap(_.value)
```
lefou, line 120: You moved the `commandChunks` upwards and made it a `def`, to make it accessible. `commandChunks` is used twice here, so maybe, we still want to hold it in a local `val`?

Flowdalic: Well, it was as val in the initial version of this PR. See https://github.com/com-lihaoyi/os-lib/pull/100#discussion_r811763520 :)

lefou: I know. But in a different position. Here, it's local and private, whereas initially you create a public val which hold redundant data. Anyway. I don't think it will hurt as-is. I'm just pointing it out. Nitpicking it is, sure, but what else can I do, if nobody else is picking up this PR.

Flowdalic: Made `commandChunks` a `val` again.

Flowdalic: Made `commandChunks` a `val` again.

PR #101, opened 2022-03-29
Add os.Path.of()

Flowdalic:
This is the second iteration after https://github.com/com-lihaoyi/os-lib/pull/88. I still feel like os-lib is lacking such a convenience method. Happy to add unit tests if you feel that this should go forward.

OS-Lib was lacking a convenience methods that lets user construct
absolute paths from a wide range of inputs values. In particular,
there is no function that takes either an absolute path, relative path
or a tilde prefixed path and constructs an absolute os.Path from it.

The newly added Path.of() is exactly such a function, plus it includes
the functionality of the already existing Path.expandUser() function.

### DIFF
--- a/os/src-jvm/package.scala
+++ b/os/src-jvm/package.scala
@@ -21,7 +21,7 @@ package object os{
   /**
    * The current working directory for this process.
    */
-  val pwd: Path = os.Path(java.nio.file.Paths.get(".").toAbsolutePath)
+  val pwd: Path = os.Path.cwd
 
   val up: RelPath = RelPath.up
 
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -374,6 +374,17 @@ object Path {
     case p: Path => p
   }
 
+  private[os] val cwd = os.Path(java.nio.file.Paths.get(".").toAbsolutePath)
+
+  /**
+    * Converts absolute paths, relative paths and automatically
+    * expands leading `~/` into the user's home directory.
+    */
+  def of[T: PathConvertible](f: T, base: Path = null) =
+    // Note that we can't use os.pwd as it would cause a cyclic
+    // dependency. Hence the package-scoped os.Path.getPwd.
+    expandUser(f, if (base == null) cwd else base)
+
   /**
     * Equivalent to [[os.Path.apply]], but automatically expands a
     * leading `~/` into the user's home directory, for convenience


Flowdalic:
Exactly. The idea is to enable the user to input *any* sensible string that represents a path to get hold of a os.Path instance. This was one of the first functions I searched for when I started using os-lib but could not find it immediately, as it is not really clear from the documentation, or even from the code itself, as we have established in our previous discussions, that `os.Path.expandUser(somePath, os.pwd)` exhibits this desired behavior.

lefou:
The point is, we already have two ways of creating a `Path`:

* `os.Path.apply`
* `os.Path.expandUser` 

Both accept various input data types via `PathConvertible`.
Both support a second parameter `base` to specify the base to use, in case the given path is relative.
The first is the default way to create a `Path` , as using the `apply` mehtod is very Scala idiomatic.
The second has extended functionality, which is expressed in it's name.

I my opinion, a third method which has neither a descriptive name nor adds new functionality isn't worth it. I'm pretty sure, a user will not understand the difference between `apply` and `of` without inspecting the documentation. Once (s)he reads the documentation, we could just document this use case there.

Please note, that I'm not the main maintainer of os-lib and may not have the final word on this.

Flowdalic:
Right, I believe the `expandUser` interface was a mistake. First, its default argument for the `base` parameter is `null`, when it should be `pwd`, as I argue that is what you most of the time want. Second, most strings I want to convert to `os.Path` are either one of an absolute path, an relative or a tilde prefixed path. So I have code like

```
val inputPaths = Seq("~/path", "path", "/path")
```

which I want to convert to os.Path. Currently I need to map this over `s => os.Path.expandUser(s, os.pwd)` which nearly gives the impression that it will either expand use or assume an absolute input path, if you do not start wondering about its second argument when reading the code.

`os.Path.of()` does not take a second argument. `Path.of` is also what the Java runtime library uses, hence users are familiar with it, and will ideally start looking for such a method right away. The method name `of` is short, indicating that it's an important method, as if you don't really care what and how to convert a String to a Path, it's the one method that will do *always* the right thing. Compare this to mapping `inputPaths` above over `s => os.Path.expandUser(s)`: Boom `IllegalArgumentException`.

lefou:
I'd like to have @lihaoyi opinion here.

IMHO, using "~" is a very shell thing, which otherwise don't crosses the way. E.g. on Windows, a "~" most likely means some kind of temporary file, but I only remember to have seen it used as prefix or suffix. That's also why I don't see the need to have this new `of` method.

Could it be, that what we really want is change the default value of `base` to `os.pwd`?

lihaoyi:
I can see the value in this. I do agree that `os.Path.expandUser(somePath, os.pwd)` is a bit verbose and roundabout, and I do agree that this is a relatively common use case we want to make more convenient.

On the other hand, `.of` is definitely far too short for what this does, and too meaningless for the specific use case that this is targetted at.

How about something like `os.Path(somePath, os.pwd, expandUser = true)`? That should be discoverable, as it can be found by anyone looking at `os.Path.apply`. We could also add an `os.FilePath(somePath, expandUser = true)` function to preserve the distinction between relative and absolute paths, for use cases that need that distinction.

Flowdalic:
> On the other hand, `.of` is definitely far too short for what this does, and too meaningless for the specific use case that this is targetted at.

Personally, I would reach to the opposite conclusion. The proposed `os.Path.of()` is **the** no-brainer method to transform a `String` to `os.Path`, which is, in my experience, a very common operation. And a good guideline for API design is that often used, popular methods should have short names. I am also not sure about the "too meaningless"  criticism. It is what Java's standard library does with [`Path.of(String)`](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/nio/file/Path.html#of(java.lang.String,java.lang.String...)). But But it would definitely be more Scala-idiomatic if the exact functionality of the proposed `os.Path.of()` method would been provided by `apply()` in `os.Path`s companion object.


> How about something like `os.Path(somePath, os.pwd, expandUser = true)`?

I am not sure if I understand the semantic of the second parameter? Is it like the existing optional relative `base` parameter that some methods in `os.Path` already have? Does it have a default value of `os.pwd`?


> We could also add an `os.FilePath(somePath, expandUser = true)` function to preserve the distinction between relative and absolute paths, for use cases that need that distinction.

Sure, as long as there is an API that does not make that distinction too. But where would one then but this functionality?

To be frank, I am not sure if I understand why `os.Path.apply()` needs to throw an IllegalArgumentException if an relative path is passed. But that's probably my being biased by command line interfaces and shells that usually relativize a relative path to the current working directory. But if you want to keep this behavior, then where shall one put the no-brainer "enter string here and see if you get a `os.Path` out of it" method? Maybe as `os.Path.of()`? ;)

PR #102, opened 2022-04-17
[fix] ctime, atime of StatInfo

mr-cloud:
fix order of ctime, atime of StatInfo and clarify the docs


lefou (approved):
Nice catch. Looks good to me. Thanks!

PR #103, opened 2022-05-13
Adjust os.Path.apply() to take relative paths and expand home

Flowdalic:
OS-Lib was lacking a convenience methods that lets user construct
absolute paths from a wide range of inputs values. In particular,
there is no function that takes either an absolute path, relative path
or a tilde prefixed path and constructs an absolute os.Path from
it.

The new behavior of os.Path.apply() accommodates for this.

Although I am not sure if I understand the proposal made in https://github.com/com-lihaoyi/os-lib/pull/101#issuecomment-1112731311, this hopefully satisfies the proposal.

### DIFF
--- a/os/src-jvm/package.scala
+++ b/os/src-jvm/package.scala
@@ -23,7 +23,7 @@ package object os{
   /**
    * The current working directory for this process.
    */
-  val pwd: Path = os.Path(java.nio.file.Paths.get(".").toAbsolutePath)
+  val pwd: Path = os.Path.cwd
 
   val up: RelPath = RelPath.up
 
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -376,6 +376,8 @@ object Path {
     case p: Path => p
   }
 
+  private[os] val cwd = new Path(java.nio.file.Paths.get(".").toAbsolutePath.normalize)
+
   /**
     * Equivalent to [[os.Path.apply]], but automatically expands a
     * leading `~/` into the user's home directory, for convenience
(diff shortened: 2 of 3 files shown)


lefou (changes requested):
> Although I am not sure if I understand the proposal made in [#101 (comment)](https://github.com/com-lihaoyi/os-lib/pull/101#issuecomment-1112731311), this hopefully satisfies the proposal.

This PR does not match the linked proposal.

Your added `apply` method is lacking the `base: Path` parameter, blurring the line between absolute and relative paths.

lefou:
> > Although I am not sure if I understand the proposal made in [#101 (comment)](https://github.com/com-lihaoyi/os-lib/pull/101#issuecomment-1112731311), this hopefully satisfies the proposal.
> 
> This PR does not match the linked proposal.
> 
> Your added `apply` method is lacking the `base: Path` parameter, blurring the line between absolute and relative paths.

This was probably a bit fast, sorry.

What I'm trying to say. This changes does not provide very useful API, a user is still forced to handle an exception in case the path is relative and does not contain a user marker. In the previous discussions, your main argument was "convenience", but this change doesn't improve on that.

Either create a `def apply(f0: PathConvertible, base: os.Path, expandUser: Boolean): Path` in `Path` to give enough options to handle relative paths, or add a `def apply(f0: PathConvertible, expandUser: Boolean)` in `FilePath` which may return a `Path` or `RelPath`.

os/src-jvm/package.scala
```diff
@@ -16,7 +16,7 @@ package object os{
   /**
    * The user's home directory
    */
-  val home: Path = Path(System.getProperty("user.home"))
+  val home: Path = os.Path.cwd
```
lefou, line 19: This is wrong, at least semantically.

os/src/Path.scala
```diff
@@ -388,14 +390,23 @@ object Path {
...
   def apply[T: PathConvertible](f: T, base: Path): Path = apply(FilePath(f), base)
-  def apply[T: PathConvertible](f0: T): Path = {
-    val f = implicitly[PathConvertible[T]].apply(f0)
-    if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount/ 2) {
-      throw PathError.AbsolutePathOutsideRoot
+  def apply[T: PathConvertible](f0: T, expandHome: Boolean = true): Path = {
+    val f = {
+      val f = implicitly[PathConvertible[T]].apply(f0)
+      if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount/ 2) {
+        throw PathError.AbsolutePathOutsideRoot
```
lefou, line 397: This will throw, when the path contains more `..` elements than path segements, which can be true for relative paths.

lefou (commented):
I outlined some concrete issues.

Flowdalic:
Thanks for your feedback.

> In the previous discussions, your main argument was "convenience", but this change doesn't improve on that.

I think it does. `os.Path.apply()` now accepts relative and absolute `PathConvertible`s as input and now is a convenient method to construct `os.Path` from nearly arbitrary inputs.

But as our [gitter discussion](https://gitter.im/lihaoyi/os-lib?at=6319b3b9aa09177429fc9562) showed, the library maintainer may not want such a convince method, neither by adjust `os.Path.apply()`, nor by keeping the semantics of `os.Path.apply()` and adding a new method (e.g., `os.Path.of()`) with the desired semantics. In that case, I failed to convince the maintainer that such a method would be a sensible addition to os-libs API and this PR can be closed.

lihaoyi:
I hate to be the bad guy here closing this PR, but we have already discussed this topic at length, and the basic changes proposed here have been rejected several times before with detailed reasoning.

Flowdalic:
> I hate to be the bad guy

IMHO that's not being the bad guy :)

> but we have already discussed this topic at length, and the basic changes proposed here have been rejected several times before with detailed reasoning.

It wasn't clear to me from the previous discussions that you absolutely resent the idea of having a convenience method to construct instances of `os.Path` that accepts absolute, relative, and tilde-prefix `PathConvertibles` as input. It is clear to me now, although I don't agree with the reasoning.

PR #104, opened 2022-05-27
Fix compilation warnings

sake92:


os/src/ReadWriteOps.scala
```diff
@@ -327,9 +327,9 @@ object read extends Function1[ReadablePath, String]{
     * can override by specifying a `charSet`.
     */
   object lines extends Function1[ReadablePath, IndexedSeq[String]]{
-    def apply(src: ReadablePath) = stream(src).toArray[String]
+    def apply(src: ReadablePath) = stream(src).toArray[String].toIndexedSeq
```
lefou, line 330: Return type?

os/test/src-jvm/ExampleTests.scala
```diff
@@ -271,7 +271,7 @@ object ExampleTests extends TestSuite{
             .filter(_._2.length > 0)
             .filter(!_._1.segments.contains("src_managed"))
 
-      assert(filesWithTooLongLines.length == 0)
+      Predef.assert(filesWithTooLongLines.length == 0, filesWithTooLongLines)
```
lefou, line 274: What's the motivation? Wasn't it utest `assert` berfore, which is printing the `filesWithTooLongLines` automatically in case of failure?

lefou (commented):
Fixing warnings is great! I added some minor comments/questions.

os/test/src-jvm/ExampleTests.scala
sake92: It wasnt printing *which files* had issues, now it does.

os/src/ReadWriteOps.scala
sake92: added

os/src/ListOps.scala
```diff
@@ -15,12 +15,12 @@ import java.nio.file.attribute.BasicFileAttributes
   * them. You can disable sorted by passing in the flag `sort = false`.
   */
 object list extends Function1[Path, IndexedSeq[Path]] {
-  def apply(src: Path, sort: Boolean = true) = {
+  def apply(src: Path, sort: Boolean = true): Array[Path] = {
```
lefou, line 18: Shouldn't this one also return a `IndexedSeq[Path]`? The other `apply` and the inherited `Functiona1` suggests it.

lefou (approved):
Looks good to me. Thank you!

PR #106, opened 2022-08-04
Make Shellable a sequence of CharSequence(s)

Flowdalic:
This allows to pass everythign that is a CharSequence to
os.proc(). Previously one had to add an extra toString() call, e.g.,:

    val host = org.minidns.dnsname.DnsName.from("example.org")
    os.proc("ping", host.toString).call()

now

    val host = org.minidns.dnsname.DnsName.from("example.org")
    os.proc("ping", host).call()


Flowdalic:
> Isn't the newly introduced `CharSequenceShellable` already enough to accept `CharSequences`?

It is. I've adjusted the PR accordingly.

lefou (approved):
Looks good to me. Thank you!

PR #108, opened 2022-08-31
Wrap the invocation of Process.waitFor() into blocking

Flowdalic:
Awaiting a child process is a potentially blocking operation and hence
should be wrapped into scala.concurrent.blocking.

### DIFF
--- a/os/src-jvm/SubProcess.scala
+++ b/os/src-jvm/SubProcess.scala
@@ -3,6 +3,7 @@ package os
 import java.io._
 import java.util.concurrent.TimeUnit
 
+import scala.concurrent.blocking
 import scala.language.implicitConversions
 
 /**
@@ -50,7 +51,7 @@ class SubProcess(val wrapped: java.lang.Process,
     * indefinitely. Returns `true` if the subprocess has terminated by the time
     * this method returns.
     */
-  def waitFor(timeout: Long = -1): Boolean = {
+  def waitFor(timeout: Long = -1): Boolean = blocking {
     if (timeout == -1) {
       wrapped.waitFor()
       true


alexarchambault:
Isn't it more the user's responsibility to wrap things in `blocking`, rather than the library's authors? Basically all os-lib methods' implementations would need to be wrapped in `blocking` else...

Flowdalic:
> Isn't it more the user's responsibility to wrap things in `blocking`, rather than the library's authors? Basically all os-lib methods' implementations would need to be wrapped in `blocking` else...

Yes, I can see how this could become cumbersome. However, there is also can argument to be made to have the `blocking` block at the lowest-possible layer, so that is as short as possible (similar how critical sections should be kept short). Personally, I would want to have it that way. Also, sometimes, the required amount of `blocking` blocks can be surprisingly few. For example, this is the only part where os-lib uses "lower" API layers to wait for a child process. I also believe that not all potential blocking call sites should be wrapped in `blocking`. For example, `stat()` calls could probably go without. There is probably a sweet spot between the two extremes of wrapping everything in `blocking` and wrapping nothing in `blocking`. And waiting for child processes appears to be a sensible case which should be wrapped, as it potentially blocks an extended period of time and could easily lead to deadlocks otherwise.

lefou:
I have two issues with this PR.

* It suggests its reader, the whole code base was written with concurrency and properly blocking demarcation in mind. This is not the case. We clearly send a false signal.

* `blocking` is discouraged and can be troublesome when used in `Future`s (see [here](https://docs.scala-lang.org/overviews/core/futures.html#blocking-inside-a-future)). But we have no way to control whether this code is used inside a Future.

It is to my knowledge non-standard for libraries to place `blocking` here and there. Instead, these kind of control statements and tweaks should be used on the users side when appropriate, to better control the behavior of the application.

lihaoyi:
I generally agree with @lefou here

Issue #111, opened 2022-09-07
Add MiMa checks to ensure binary compatibility


lefou:
We can use [mill-mima](https://github.com/lolgab/mill-mima) plugin for it.

PR #115, opened 2022-09-08
Enable support for MiMa checks

lefou:
Turns out, we already broke our binary API since 0.8.0 in two pull requests.

So, I will de-facto disable the checks as we have no version to which we want to stay compatible.
We can fill the list, once we make the next release.

Also added GHA CI action for MiMa checks.

Here are the PRs which broke bin compat:
* #100 
* #104 

I don't think we should revert them. 
To the contrary, both contain useful improvements.
Instead, we should try to improve the code base even more and probably make changes which break the API now.
After that we should create a test release 0.9, and when we find it good, also create 1.0 release.

* Implements #111

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -35,6 +35,20 @@ jobs:
       - run: ./mill.bat -i -k __.jvm.__.test
         if: matrix.os == 'windows-latest'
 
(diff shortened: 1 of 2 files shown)


lolgab (approved):
LGTM! 👍

Issue #117, opened 2022-09-11
ConcurrentLinkedQueue issues with scala native


alterationx10:
I have a project where I'm trying to use os-lib 0.8.1 (and scala native 0.4.7). If I try to run my code, even with os-lib just in the imports, I will get:

```
[error] Found 5 missing definitions while linking
[error] Not found Member(Top(java.util.concurrent.ConcurrentLinkedQueue),D3addL16java.lang.ObjectzEO)
[error]         at /home/runner/work/zio/zio/core/shared/src/main/scala/zio/internal/FiberRuntime.scala:1253
[error] Not found Member(Top(java.util.concurrent.ConcurrentLinkedQueue),D4pollL16java.lang.ObjectEO)
[error]         at /home/runner/work/zio/zio/core/shared/src/main/scala/zio/internal/FiberRuntime.scala:223
[error] Not found Member(Top(java.util.concurrent.ConcurrentLinkedQueue),D5offerL16java.lang.ObjectzEO)
[error]         at /home/runner/work/zio/zio/core/shared/src/main/scala/zio/internal/impls/LinkedQueue.scala:46
[error] Not found Member(Top(java.util.concurrent.ConcurrentLinkedQueue),D6addAllL20java.util.CollectionzEO)
[error]         at /home/runner/work/zio/zio/core/shared/src/main/scala/zio/internal/impls/LinkedQueue.scala:53
[error] Not found Member(Top(java.util.concurrent.ConcurrentLinkedQueue),D7isEmptyzEO)
[error]         at /home/runner/work/zio/zio/core/shared/src/main/scala/zio/internal/FiberRuntime.scala:222
```

That ZIO source code above calls > `private val queue   = new java.util.concurrent.ConcurrentLinkedQueue[FiberMessage]()`

I see that there is an implementation of ConcurrentLinkedQueue in os-lib, but there is also a version supplied by scala-native. 

https://github.com/scala-native/scala-native/blob/main/javalib/src/main/scala/java/util/concurrent/ConcurrentLinkedQueue.scala

I was able to build a local version with the os-lib implementation removed, and using that, my project worked just fine. Could to os-lib one be safely removed from the project?

lefou:
Fixed by #122

PR #122, opened 2022-09-18
Delete `ConcurrentLinkedQueue`

armanbilge:
This is now in Scala Native javalib.


lefou:
It seems this fixes https://github.com/com-lihaoyi/os-lib/issues/117 and it looks reasonable to me. Although, I have no idea at which point in time this class was added to Scala native directly, and whether there is a chance to break some users code. Our binary checks suggest, it won't, at least not for the same set of versions. Maybe @lolgab can review this one, as he initially added this class.

armanbilge:
Great point. This has been there since Scala Native 0.4.0.

https://github.com/scala-native/scala-native/issues/1993

Since that was a breaking release with entirely new artifacts I think we can be confident that this won't break anybody.

lefou:
Thanks @armanbilge. Looks good to merge then.

lolgab:
Thank you for the fix @armanbilge and for merging @lefou.

PR #128, opened 2022-10-25
Converted Readme to Asciidoc, reviewed and fixed links and typos

lefou:
Asciidoc(tor) supports automatic table of contents, so I deleted the hand-crufted one. Also replaced all occurrences of a version with a `version` attribute, so we now only need to update the attributes in the document header. I chose `toc-placement` `preamble` which places the TOC after the preable but before the first section, which is exactly where the old TOC was.

Also updated some code examples and replaced deprecated API calls. I think there are still some outdated parts in the documentation, but this needs to be addressed separately.


jodersky (approved):
thanks

PR #129, opened 2022-10-26
Enforce safe (transitive) Scala library

lefou:
We do this to avoid known security vulnerabilities of older Scala library versions.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -116,7 +116,7 @@ trait SafeDeps extends ScalaModule {
   override def mapDependencies: Task[coursier.Dependency => coursier.Dependency] = T.task {
     val sd = Deps.scalaLibrary(scala213Version)
     super.mapDependencies().andThen { d =>
-      // enforce up-to-date Scala 3.13 version
+      // enforce up-to-date Scala 2.13.x version
       if (d.module == sd.dep.module && d.version.startsWith("2.13.")) {
         sd.dep
       } else d
--- a/build.sc
+++ b/build.sc
@@ -4,7 +4,7 @@ import $ivy.`com.github.lolgab::mill-mima::0.0.13`
 
 // imports
 import mill._
-import mill.define.Target
+import mill.define.{Task, Target}
 import mill.scalalib._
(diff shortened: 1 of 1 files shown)


SethTisue:
I haven't seen anything like this done in other repos — I'm curious if you'd care to explain what the thinking here is.

lefou:
We enforce some versions so we don't use older ones transitively. This is mostly relevant for tests here, but it also makes Dependabot's security scanner happy, as it is currently inspecting all dependencies, including runtime and test-scoped ones.

lefou:
Also relevant, the lack of a better management mechanisms in Mill (and Coursier), which will also take effect when the library is used transitively.

* https://github.com/com-lihaoyi/mill/issues/1975
* https://github.com/coursier/coursier/issues/1390

SethTisue:
cool, thanks

build.sc
```diff
@@ -93,7 +96,35 @@ object os extends Module {
...
+  override def compileIvyDeps = acyclicDep
+  override def scalacPluginIvyDeps = acyclicDep
+  override def scalacOptions = T {
+    super.scalacOptions() ++ acyclicOptions()
+  }
+}
+
+trait SafeDeps extends ScalaModule {
+  override def mapDependencies: Task[coursier.Dependency => coursier.Dependency] = T.task {
+    val sd = Deps.scalaLibrary(scala213Version)
```
alexarchambault, line 117: Why not simply add this dependency when `scalaVersion().startsWith("2.13.")`? That would ensure the Scala 2.13 version is `>= scala213Version`.

lefou: This is not working as it is a special case. `ScalaModule.mapDependency` replaces all occurences of a `scala-library` by the assumed correct version matching the `scalaVersion`, which can be either `scala-library` or `dotty-library`  or `scala3-library` and friends. 

Effectively, I'm only fixing the transitive older scala-library that comes in via Scala `3.1.3`. In case I build for Scala `2.13.10`, this fix isn't effective and not needed.

PR #139, opened 2022-12-05
Fix watcher test for OSX

dm3:
For some reason a part of the watcher test for OSX which has been commented out before is working now.

Tested on OSX 15, Apple M1


lefou (approved):
Looks good to me. Thank you!

PR #142, opened 2022-12-21
Add a `PathConvertible` for `URI`

ckipp01:
I needed to use this a couple times in a project recently so I figured I'd try to upstream it.

### DIFF
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -1,5 +1,8 @@
 package os
 
+import java.net.URI
+import java.nio.file.Paths
+
 import collection.JavaConverters._
 import scala.language.implicitConversions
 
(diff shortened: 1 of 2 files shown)


lefou (commented):
As not all `URI`s can be converted to `Path`s, this PR should also include a test case covering an error situation, which probably means to assert an thrown `IllegalArgumentException`.

os/src/Path.scala
```diff
@@ -501,12 +504,21 @@ sealed trait PathConvertible[T] {
...
   }
   implicit object NioPathConvertible extends PathConvertible[java.nio.file.Path] {
     def apply(t: java.nio.file.Path) = t
   }
+  implicit object UriPathConvertible extends PathConvertible[URI] {
+    def apply(uri: URI) = uri.getScheme() match {
+      case "file" => Paths.get(uri)
+      case uriType =>
+        throw new IllegalArgumentException(
+          s"""os.Path can only be created from a "file" URI scheme, but found ${uriType}"""
```
lefou, line 520: I would also quote the found URI type.

lefou (commented):
One minor thing below, otherwise this looks good to me.

os/src/Path.scala
ckipp01: Sure, fixed!

PR #145, opened 2022-12-21
Enable Scalafmt and format everything

ckipp01:
This PR adds in 2 commits. One being a formatting commit since there is a `.scalafmt.conf` in this repo, but it was never enforced. The second commit adds a check in CI for formatting and also creates a `.git-blame-ignore-revs` file with the formatting commit in it.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,2 @@
+# Format the codebase for the first time
+b049841d5707d5bd87be516d8cda7be2a7585eae
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -49,6 +49,20 @@ jobs:
 
       - run: ./mill -i -k __.mimaReportBinaryIssues
 
+  check-formatting:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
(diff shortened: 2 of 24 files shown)


lefou (changes requested):
Can you please isolate the formatting commit even more. I'd like to have the tooling changes to enable the `ScalafmtModule` in `build.sc` not end up in `.git-blame-ignore-revs`.

ckipp01:
> Can you please isolate the formatting commit even more. I'd like to have the tooling changes to enable the `ScalafmtModule` in `build.sc` not end up in `.git-blame-ignore-revs`.

Sure, no problem. I've adjusted it now so that the commit that is ignored _only_ has the actual formatting changes.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Format the codebase for the first time
+6ae7a064eb53980dd391f6bd6484c12efe7083b4
```
lefou, line 2:
```suggestion
b049841d5707d5bd87be516d8cda7be2a7585eae
```

lefou (changes requested):
The `.git-blame-ignore-revs` still references an outdated commit. Beside that, this PR looks good to me.

.git-blame-ignore-revs
ckipp01: Ah dangit, totally forgot to update it after I rebased. Thanks! Fixed.

lefou (approved):
Looks good to me. Thank you!

PR #146, opened 2022-12-26
fix: handle `baseName` for hidden dirs

ckipp01:
I had a usecase for this when I was working with `.well-known` as a dir
and got bite by this since `baseName` on this will just return `""`.
This changes the logic slightly to return `last` if either `-1` is found
meaning there is no `.` in the file, or if the index is `0` indicating
that it's hidden and there is also no extension. I've also added a test
for this.

### DIFF
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -200,7 +200,8 @@ trait BasePathImpl extends BasePath {
 
   override def baseName: String = {
     val li = last.lastIndexOf('.')
-    if (li == -1) last
+    // We also check 0 for things like hidden files/directories that start with .
+    if (li == -1 || li == 0) last
     else last.slice(0, li)
   }
 
--- a/os/test/src/PathTests.scala
+++ b/os/test/src/PathTests.scala
@@ -54,6 +54,7 @@ object PathTests extends TestSuite{
           assert((base / "baseName.v2.0.ext").baseName == "baseName.v2.0")
           assert((base / "baseOnly").baseName == "baseOnly")
           assert((base / "baseOnly.").baseName == "baseOnly")
+          assert((base / ".well-known").baseName == ".well-known")
         }
 
         test("ext"){


lefou (commented):
This is a tricky one. At least, I can imagine there are different opinions about that. From a UNIX perspective, `.well-known` is an _hidden_ file with the name `well-known` (prefixed with a dot to make it hidden). From a Windows perspective, this is a ordinary file (or dir) with no basename but the extension `well-known`. Before merging, gathering more opinions might be nice, links to how other file system abstractions handle that case would be even better. 

In any case, we need to add some documentation to the `ext` `def` as well as the documentation itself, if we plan to special case a leading dot in the name. Currently `os-lib` does not abstract over the concept of _hidden_ files at all.

lefou:
> I had a usecase for this when I was working with `.well-known` as a dir and got bite by this since `baseName` on this will just return `""`. 

Why did you use the `basename` in this case. Wouldn't just `last` be the better options? Otherwise you risk to also catch wrong or multiple files like  `.well-known.bak` and `.well-known.0`.

ckipp01:
Good questions.

> This is a tricky one. At least, I can imagine there are different opinions about that. From a UNIX perspective, .well-known is an hidden file with the name well-known (prefixed with a dot to make it hidden).

Yes, I'm aware of this, but if it's hidden is it supposed to change what `basename` returns? That's what wasn't clear to me looking through the code as it doesn't seem like hidden files are really special cased at all.

> Why did you use the basename in this case. Wouldn't just last be the better option? Otherwise you risk to also catch wrong or multiple files like .well-known.bak and .well-known.0.

I honestly didn't think of `last`. Yea that would work better for my usecase. However, your example with `.bak` illustrates why I think this chnage is an important one. I think it's confusing if `.well-known` and `.well-know.bak` have different basnames. I would have assumed that in both of those scenarions, `.well-known` would have been returned.

lefou:
> > Why did you use the basename in this case. Wouldn't just last be the better option? Otherwise you risk to also catch wrong or multiple files like .well-known.bak and .well-known.0.
> 
> I honestly didn't think of `last`. Yea that would work better for my usecase. However, your example with `.bak` illustrates why I think this chnage is an important one. I think it's confusing if `.well-known` and `.well-know.bak` have different basnames. I would have assumed that in both of those scenarions, `.well-known` would have been returned.

I depends on whether we settle with the interpretation of _hidden_ files. I for myself need to find out about the platform I don't know so well. E.g. Does Windows support files with empty basesames? Are dot-files handled like files with empty basename and full extension? 

Please note, that our current implementation of `.ext` is pure. It's not using any kind of system specific knowledge like reading `/etc/mime.types` and stuff like this. A `archive.tar.gz` will have an extension of `gz`, not `tar.gz`. I think we should try to avoid different behavior based on the executing platform. If we find out, Windows is interpreting dot-files as files with basename without extension, we can just proceed. Otherwise, we're probably better off by introducing a new `def`.

Issue #159, opened 2023-03-22
Leading `os.up` in a Seq used in a path segment are not properly handled


lihaoyi:
```scala
@ os.rel / Seq(os.up, os.rel/"hello", os.rel/"world")
res0: os.package.rel.ThisType = hello/world
```

I would expect this to print `../hello/world`, but the leading `..` is improperly dropped

lefou:
Good catch! Same happens with `Array`s. 

```scala
@ os.rel / Array(os.up, os.rel/"hello", os.rel/"world")
res0: os.package.rel.ThisType = hello/world
```

PR #160, opened 2023-03-24
Fix `SeqPathChunk` 

lihaoyi:
Fixes https://github.com/com-lihaoyi/os-lib/issues/159

### DIFF
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -41,10 +41,12 @@ object PathChunk {
     override def toString() = inner.toString
   }
   implicit class SeqPathChunk[T](a: Seq[T])(implicit f: T => PathChunk) extends PathChunk {
+    @deprecated("never used, really shouldn't exist, kept for bincompat")
     var segments0 = Nil
+    @deprecated("never used, really shouldn't exist, kept for bincompat")
     var ups0 = 0
 
-    val rel = a.map(f).foldLeft(os.rel) { case (current, chunk) => current / chunk}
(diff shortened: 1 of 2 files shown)


os/src/Path.scala
```diff
@@ -43,9 +43,9 @@ object PathChunk {
   implicit class SeqPathChunk[T](a: Seq[T])(implicit f: T => PathChunk) extends PathChunk {
     var segments0 = Nil
     var ups0 = 0
```
lefou, line 45: Any idea, why these are here? Could you add some `@deprecated` on them, so we can remove them in the next major bump?

os/test/src/PathTests.scala
```diff
@@ -425,5 +425,10 @@ object PathTests extends TestSuite{
         }
       }
     }
+    test("issue159"){
+      val result = os.rel / Seq(os.up, os.rel/"hello", os.rel/"world")
+      val expected = os.up / "hello" / "world"
+      assert(result == expected)
```
lefou, line 431: Would be nice to also cover the `ArrayPathChunk` case here, as it's also affected by #159.

lihaoyi:
@lefou do you know what to do with these `os.jvm[3.1.3].checkFormat Found 1 misformatted files` errors? Is there some command I can run to autoformat things compliantly?

lefou:
> @lefou do you know what to do with these `os.jvm[3.1.3].checkFormat Found 1 misformatted files` errors? Is there some command I can run to autoformat things compliantly?

Yeah. You can run:

```
$ mill __.reformat
```

Issue #47, opened 2020-09-28
Publish Scala 2.11 builds again?


tuxji:
I am working on an Apache Daffodil pull request (still in progress in my own fork) and I am using os-lib to write files and run executables on these files.  Everything was working fine until I enabled Daffodil's GitHub Actions CI workflow in my fork.  It turns out that the CI workflow cross-compiles Daffodil under both Scala 2.12 and 2.11 (the Daffodil committers haven't added support for Scala 2.13 yet).  I am using os-lib v0.7.1 and the last time os-lib was published for Scala 2.11 was os-lib v0.2.9 in March 2019.  Since os-lib depends on geny v0.6.2 and geny also stopped publishing Scala 2.11 builds with v0.1.6 in January 2019, a geny build would be needed too.  

Would you be willing to publish both geny and os-lib Scala 2.11 builds again for a period of time?  Until Daffodil stops supporting Scala 2.11?  Is it technically possible to do that?

tuxji:
The Daffodil devs agreed it was time to stop making Scala 2.11 builds, so this is no longer an issue.

PR #196, opened 2023-08-07
Add root to Path & root constructor

### LINKED ISSUE #170: Add optional root in Path and support for roots for Windows
Per the following discussion: https://github.com/com-lihaoyi/os-lib/discussions/136
I propose that we: 
 * Introduce a getter for optional root in `Path`
 * Return the `root` in `segments`. Java's `Path` returns segments without the drive letter, but keeps the information. I.e. - the path is relative, root is not present, and the .getRoot on segments return null, but the segments can be brought back to be absolute with `.toAbsolutePath`. `segments` in os-lib do something different that Java's Path iterator, as it returns only Strings, not relative paths that can be made absolute.
 * TBD: Add `RootPath` class to expose `apply` accepting a drive letter. It could be the type of `os.root`, so people could do either `os.root / ...` or `os.root("C:") / ...`. We could also add `os.drive("C:")` or something like that, and leave the `os.root` defaulting to the system drive.

szymon-rd:
Resolve #170 

This PR adds a `def root(root: String, fileSystem: FileSystem): Path` in `os` package object. Additionally, it adds `root` and `filesystem` members to `Path`. It addresses two problems:
 - Specifying custom roots for a path
 - Using custom filesystems

## Filesystem

Filesystem was added as a field of the `os.Path`. It was not added as a Root field for simplicity and consistency with Java's `nio`. Root is a part of the path; filesystem is the context of the whole path. One filesystem can have many roots; one root can have many subdirectories.

## Root as String

A string is the simplest type that can represent the path's root. It also corresponds to the mental model of the path - usually, developers perceive it as a String. Therefore - String was chosen as the type of root.

Pull request: https://github.com/com-lihaoyi/os-lib/pull/196


lihaoyi:
I think this looks reasonable. Is it possible to make the change binary compatible? Something like

```scala
val root: Path = ...
def root(s: String): Path = ...
```

szymon-rd:
@lihaoyi seems to work all right

os/src/Path.scala
```diff
@@ -484,6 +484,7 @@ class Path private[os] (val wrapped: java.nio.file.Path)
     new SeekableSource.ChannelSource(java.nio.file.Files.newByteChannel(wrapped))
 
   require(wrapped.isAbsolute || Path.driveRelative(wrapped), s"$wrapped is not an absolute path")
+  def root = Option(wrapped.getRoot).map(_.toString).getOrElse("")
```
lihaoyi, line 487: Does root have to be a string? I wonder if we can make it a `java.nio.file.Path`, or some opaque wrapper of one `trait os.RootPath{ def toString: String }` or something. 

We could keep the `def root(String)` factory method for convenience, but making the `def root` somewhat less concrete would give us flexibility in future if we find we want to expose more of the underlying `java.nio` capabilities

szymon-rd: I am not sure. If we knew how it could be used, we could leave that open, but putting just the String there seems to be the simplest way to go. The Path needs additional checks, and I am not sure what it would mean - most of the Paths are not just roots, and even if they are, one is able to construct os.Path from them in another way than os.root. Unless you mean extracting just the root from the nio Path?

lihaoyi: One case that was mentioned in the other thread: `java.nio.file.Path`s are bound to a specific `java.nio.file.Filesystem`. And can refer to e.g. in-memory filesystems or zip-filesystems and other things. 


```scala
lihaoyi mill$ amm
Loading...
Welcome to the Ammonite Repl 2.5.4 (Scala 2.13.8 Java 17.0.6)
@ import collection.JavaConverters._
import collection.JavaConverters._

@ val zipFs = java.nio.file.FileSystems.newFileSystem(java.net.URI.create("jar:file:///usr/local/bin/amm"), Map("create" -> "true").asJava)
zipFs: java.nio.file.FileSystem = /usr/local/bin/amm

@ java.nio.file.Paths.get("/") == java.nio.file.Paths.get("/")
res2: Boolean = true

@ zipFs.getPath("/") == java.nio.file.Paths.get("/")
res3: Boolean = false

@ zipFs.getPath("/").getFileSystem
res4: java.nio.file.FileSystem = /usr/local/bin/amm

@ java.nio.file.Paths.get("/").getFileSystem
res5: java.nio.file.FileSystem = sun.nio.fs.MacOSXFileSystem@4228bf58
```

This raises two questions:

1. How does that work on windows? Is each windows "drive" a separate `Filesystem`, or are they all on one filesystem with a different `getRoot` path?

2. Should `os.root` be able to refer to different filesystems? Given that we are right now considering how to flesh out `os.root` to make it less anemic, I feel like we might as well flesh it out "all the way". Otherwise we may end up coming back and having to repeat this whole process again later.

I feel like having `os.Path#root` return a non-string seems like the right thing to do; `java.nio.file.Path` is more complex than just a string, and we should model that somehow. Having it return `trait Root{ def prefix: String; def filesystem: java.nio.file.Filesystem }` seems like a more future-proof way of specifying it

szymon-rd: Those are fair points; I will investigate.

szymon-rd:
@lihaoyi I pushed a commit that adds a separate field `filesystem` to the Path. I'd argue that it is a separate entity from the root. I propose a model where a root can be thought of as just a string that defines the root element of a path, a filesystem can have many roots and is not a member of a root.

os/src/Path.scala
philwalk: Sounds good to me.

os/test/src-jvm/PathTestsJvmOnly.scala
```diff
@@ -40,6 +44,18 @@ object PathTestsJvmOnly extends TestSuite {
...
+          val path: java.nio.file.Path = java.nio.file.Paths.get("foo.jar");
+          val uri = new URI("jar", path.toUri().toString(), null);
+
+          val env = new HashMap[String, String]();
+          env.put("create", "true");
+
+          val fileSystem = FileSystems.newFileSystem(uri, env);
+          val p = os.root("/", fileSystem) / "test" / "dir"
+          assert(p.root == "/")
+          assert(p.fileSystem == fileSystem)
```
lihaoyi, line 57: Could we add a few more operations here? e.g. running `os.read`, `os.write`, `os.copy`, `os.move` on files with `foo.jar`.

Perhaps in that case it's worth moving this test suite into its own file

lihaoyi:
@szymon-rd looks good! Last comments on my side and then we can merge this:

1. Flesh out the test suite; left a comment there with more details. That will help cover some of the edge cases and really define what we expect to work and not work when dealing with different `root`s and different `filesystem`s
2. Need to get the Windows CI passing
3. Update the PR description with an explanation of the change: what you did, why you chose this approach over others (e.g. choosing to have a separate `filesystem` field vs having the `root` be a `java.nio.file.Path`)
4. Add a section to the readme. This could either go inside the `os.Path` section as a subsection `Roots and Filesystems`, to cover what is (a) expected to work and (b) expected to not work when people are dealing with paths on different roots and filesystems. Basically a english summary of the test suite above

szymon-rd:
@lihaoyi I think I've addressed all of your comments

lihaoyi:
Looks good to me. Thanks for working through all the feedback! It was definitely a lot of back and forth.

It seems MIMA is complaining, but I suspect the problem is Scala 2.11.x only, since the encoding of `trait`s changed between 2.11 and 2.12:

```
Found 1 issue when checking against com.lihaoyi:os-lib_2.11:0.9.0
 * method isCustomFs(java.lang.Object)Boolean in trait os.PathConvertible is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("os.PathConvertible.isCustomFs")
Found 1 issue when checking against com.lihaoyi:os-lib_2.11:0.9.1
 * method isCustomFs(java.lang.Object)Boolean in trait os.PathConvertible is present only in current version
   filter with: ProblemFilter.exclude[ReversedMissingMethodProblem]("os.PathConvertible.isCustomFs")
```

I think I'd be ok with with just adding the MIMA exclusion to shut it up, or just turning off MIMA for 2.11 entirely. 2.11 support is already pretty best-effort-ish (we removed it once in the past and added it back), so I don't feel like we should contort the code to preserve strict compatibility there just for 2.11.x. 

@lolgab @lefou let me know if you have any opinion on this, otherwise I will add a MIMA exclusion and merge this.

SethTisue:
(a voice whispers) or just drop 2.11 entirely...

SethTisue:
> 2.11 support is already pretty best-effort-ish (we removed it once in the past and added it back)

I wondered about the history on this, but I didn't find anything except #66 (which didn't come with an explanation) and #47 (which the requester withdrew).

If y'all are considering simply dropping it again, I encourage you to do so.,

lefou:
FS Root support, esp. for Windows, was my personal blocker for a 1.0 release. Now, it looks like that we can even keep binary compatibility, which is really nice. I think, we should cut a `0.9.2` after we merged it. After that, we may focus on cleaning the API and the docs for a `1.0`. In 1.0 we can drop support for Scala 2.11 and also bump our dependencies to Scala.js and Scala Native to newer versions (which we currently don't due to, again, compatibility).

lihaoyi:
Added a mima exclude, will squashmerge once green

lihaoyi:
@szymon-rd let me know if you'd like me to tag a release, or if you want to wait for your other PR to land first before tagging one

PR #200, opened 2023-08-09
Process piping

### LINKED ISSUE #187: Add utility for more concise piping
Right now, the only way to express a piping chain is something like this:
```scala
val curl = os.proc("curl", "-L" , "https://git.io/fpfTs").spawn(stderr = os.Inherit)
val gzip = os.proc("gzip", "-n").spawn(stdin = curl.stdout)
val sha = os.proc("shasum", "-a", "256").spawn(stdin = gzip.stdout)
```
I suggest that we could create additional API members that would allow expressing piping via a construct similar to the following one:
```scala
os.proc("curl", "-L" , "https://git.io/fpfTs")
   .pipeTo(os.proc("gzip", "-n"))
   .pipeTo(os.proc("shasum", "-a", "256"))
   .spawn(stderr = os.Inherit)
``` 
It could require introducing a new type, for example, a`ProcessGroup`.

szymon-rd:
Resolve #187 

This PR adds a way to pipe processes concisely and conveniently. Previously, creating pipelines behaving similarly to Unix's was impossible (or hard). `Broken pipe` error was handled by the parent process, so pipeline like `yes | head -n 5` would run forever, and the app itself crash/throw error. Additionally, there was no way to pass arguments like `pipefail`. 

All this is done with custom handler threads and not the JVM 9 API. Java's API is also incapable of handling `Broken pipe`.

On Windows, the broken pipe handling is not supported - there is no `broken pipe` error. 

Pull request: https://github.com/com-lihaoyi/os-lib/pull/200


szymon-rd:
@lihaoyi  Could you run the workflow, please? Btw, why does it require approval? Maybe it would be better for future contributors to disable that.

lefou:
Approval is a security measure. Once, you have contributed (at least one commit to git `main` branch), no further approval is needed.

lefou:
If you want to reproduce CI issues, you could also push (or create a pull request) to your own fork.

szymon-rd:
Tests are still a bit flaky on the CI, next commit should finally fix that :)

szymon-rd:
However, I think it is ready for review aside from that minor issue (@lefou @lihaoyi)

os/test/src-jvm/ProcessPipelineTests.scala
```diff
@@ -0,0 +1,163 @@
...
+object ProcessPipelineTests extends TestSuite {
+  val scriptFolder = pwd / "os" / "test" / "resources" / "scripts"
+
+  lazy val scalaHome = sys.env("SCALA_HOME")
+
+  def isWindows = System.getProperty("os.name").toLowerCase().contains("windows")
+
+  def scriptProc(name: String, args: String*): Seq[String] =
+    Seq(
+      "scala-cli" + (if (isWindows) ".bat" else ""),
```
lihaoyi, line 20: Is it possible to use a smaller more ubiquitous process in these test cases here? I don't want to have to install the whole of `scala-cli` just to run unit tests locally, especially when it seems some combination of simple unix tools (`cat`, `sleep`, `bash -c`, etc.) should be able to exercise it

Readme.adoc
```diff
@@ -1657,6 +1657,48 @@ val sha = os.proc("shasum", "-a", "256").spawn(stdin = gzip.stdout)
...
+stdin of other process in pipeline on every IO error. This behavior can be disabled via the 
+`handleBrokenPipe` flag on `call` and `spawn` methods. Note that Windows does not support 
+broken pipe behaviour, so a command like`yes` would run forever. `handleBrokenPipe` is set 
+to false by default on Windows.
+
+Both `call` and `spawn` correspond in their behavior to their counterparts in the `os.proc`,
+but `spawn` returns the `os.ProcessPipeline` instance instead. It offers the same 
+`API` as `SubProcess`, but will operate on the set of processes instead of a single one.
+
+`Pipefail`` is enabled by default, so if any of the processes in the pipeline fails, the whole
```
lihaoyi, line 1697: Extra backtick here after `Pipefail`

lihaoyi:
Some nitpicks, but otherwise looks good I think

os/test/src-jvm/ProcessPipelineTests.scala
lihaoyi: lets go with test jars then. Mill makes submodules cheap, we can easily have a few submodules to create jars for testing purposes

szymon-rd: Could you please provide an example of a decent simple workflow generating jars like this? I am not familiar with Mill, the configuration of the project you provided is quite complex, and my version will probably be a bit overcomplicated, leading to more work in the review-fixes cycle.

lihaoyi: Something like this should work

```diff
diff --git a/build.sc b/build.sc
index 80437ae..c28ab6b 100644
--- a/build.sc
+++ b/build.sc
@@ -58,6 +58,10 @@ trait MiMaChecks extends Mima {
   )
 }
 
+object testJar1 extends JavaModule
+object testJar2 extends JavaModule
+object testJar3 extends JavaModule
+
 trait OsLibModule
     extends CrossScalaModule
     with PublishModule
@@ -84,7 +88,12 @@ trait OsLibModule
     def ivyDeps = Agg(Deps.utest, Deps.sourcecode)
 
     // we check the textual output of system commands and expect it in english
-    def forkEnv = super.forkEnv() ++ Map("LC_ALL" -> "C")
+    def forkEnv = super.forkEnv() ++ Map(
+      "LC_ALL" -> "C",
+      "TEST_JAR_1_ASSEMBLY" -> testJar1.assembly().path.toString,
+      "TEST_JAR_2_ASSEMBLY" -> testJar2.assembly().path.toString,
+      "TEST_JAR_3_ASSEMBLY" -> testJar3.assembly().path.toString,
+    )
   }
 }
```

You can then put the source code for your test jars in `testJar1/src/TestJar1.java`, and use the assemblies in your test suite via `os.proc("java", "-jar", sys.env("TEST_JAR_1_ASSEMBLY"))).call()`

szymon-rd: in the tests `java.util.NoSuchElementException: key not found: TEST_JAR_READER_ASEEMBLY` happens for some reason

szymon-rd:
@lihaoyi Done I think

szymon-rd:
And - Are we going to release M1 for 1.0 now?

lefou:
> As for 1.0, I support we could. The library has been pretty stable the last 1-2 years, so once this merges it might be time to put a ribbon on it and call it stable. @lefou WDYT?

Let's cut a `0.9.2` first. I the end, we managed to stay binary compatible. Also, to get the new features / API out as soon as possible. 

For 1.0, I'd like to do some more polishing, like bumping Scala versions and dropping older Scala.js (and Scala Native) versions as a consequence. I'd rather avoid supporting one platform in 1.0 and drop it in 1.0.1, so better, we have the 0.9.2.

szymon-rd:
It's still 0.x, so semver doesn't apply here. But shouldn't it be 0.10.0? We are releasing a new, backward-compatible feature.

lefou:
> It's still 0.x, so semver doesn't apply here. But shouldn't it be 0.10.0? We are releasing a new, backward-compatible feature.

We're using early semver, which means, we handle compatibility like this: `0.major.minor`. So a `0.9.2` is perfectly fine, as it is semantically a minor release, it's adds something but does not remove anything. I'd avoid a `0.10` (a early major), as we want to release a `1.0.0` soon.

PR #202, opened 2023-08-27
fix for issue-201

### LINKED ISSUE #201: Error converting relative paths to absolute paths in Windows if relative path has leading slash or backslash
**Version(s)**
```bash
os-lib release 0.9.1
```

**Describe the bug**
The following idiom is intended to convert a relative path to an absolute path:
```scala
val abspath = Path(relativeDir, pwd)
```
On `Windows`, this idiom fails, due to there being two types of relative path, unlike on other platforms.

Windows paths are described here: [windows-file-path-formats](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats).   A reference to what might be called `rootRelative` is [here.](https://learn.microsoft.com/en-us/dotnet/standard/io/file-path-formats#:~:text=%5CProgram%20Files%5CCustom,the%20current%20drive.)

**To Reproduce**

The following code illustrates the problem with `rootRelative` paths in Windows.
```scala
Example code:

    val p1 = java.nio.file.Paths.get("/omg")
    printf("isAbsolute: %s\n", p1.isAbsolute)
    printf("%s\n", p1)
    printf("%s\n", os.Path("/omg"))
```
Output on Linux or OSX:
```sh
isAbsolute: true
/omg
/omg
```

On Windows:
```sh
isAbsolute: false
\omg
java.lang.IllegalArgumentException: requirement failed: \omg is not an absolute path
        at os.Path.<init>(Path.scala:474)
        at os.Path$.apply(Path.scala:426)
        at oslibChek$package$.main(oslibChek.sc:11)
        at oslibChek$package.main(oslibChek.sc)
```

**Expected behaviour**
A `rootRelative` path is supported by `java.nio.file.Paths.get()` but not by `os.Path()`.

In Windows, calls to `java.io.File#isAbsolute` is insufficient to determine whether a path is relative to `pwd`.
A `rootRelative` paths in `Windows` is relative to the `current working drive` (an immutable value captured at jvm startup).

There are 3 use-cases when creating a `java.nio.file.Path` object:

- path is relative - interpret path as immediately below `pwd`
- path is absolute - `java.nio.file.Paths.get(path)`
- path is `rootRelative` - interpret path as relative to the `current working drive`.

philwalk:
Thanks for a great library!   Hopefully this change will increase interest among windows shell environment developers.

This fixes #201

The essence of the fix is to support paths with a leading slash (`'/' or '\'`) on Windows.

In this PR, this path type is referred to as `driveRelative`, due to subtle differences in how it is handled on `Windows` versus other platforms.

Example code:
```scala
    val p1 = java.nio.file.Paths.get("/omg")
    printf("isAbsolute: %s\n", p1.isAbsolute)
    printf("%s\n", p1)
    printf("%s\n", os.Path("/omg"))
```
Output on Linux or OSX:
```sh
isAbsolute: true
/omg
/omg
```
On Windows:
```scala
isAbsolute: false
\omg
java.lang.IllegalArgumentException: requirement failed: \omg is not an absolute path
        at os.Path.<init>(Path.scala:474)
        at os.Path$.apply(Path.scala:426)
        at oslibChek$package$.main(oslibChek.sc:11)
        at oslibChek$package.main(oslibChek.sc)
```
### Background
On Windows, a `driveRelative` path is considered relative because a drive letter prefix is required to fully resolve it.
Like other platforms, Windows also supports `posix relative` paths, which are relative to the `current working directory`.

Because all platforms, including `Windows`, support absolute paths and `posix relative` paths, it's convenient when writing platform-independent code to view `driveRelative` paths as absolute paths, as they are on other platforms.

On Windows, the `current working drive` is an immutable value that is captured on jvm startup.  Therefore, a `driveRelative` path on Windows unambiguously refers to a unique file with a hidden drive letter.

This PR treats `driveRelative` paths as absolute paths in Windows, and has no effect on other platforms.

#### Making `os/test/src/PathTests` platform independent

This PR enables `Unix()` tests in `os/test/src/PathTests.scala`  so they also verify required semantics in Windows.

### How this PR relates to #170 and #196
The purpose of #196 seems to be to add full support on Windows, without resorting to `java.nio` classes and methods.
The addition of `os.root(...)` or `os.drive(...)` would be convenient for people writing windows-only code, whereas this PR is intended to allow writing platform-independent code, so the concerns seem to be orthogonal. 

It seems important not to alter `Path.segments` in a way that is incompatible with `java.nio` segments.
 
An alternate way to preserve drive letter information would be to add a method to one of the `Path.scala` traits that returns a drive letter if appropriate on `Windows` and an empty String elsewhere.

```scala
trait PathChunk {
  def segments: Seq[String]
  def rootPrefix: String  // empty String, or Windows drive letter
  def ups: Int
}
```

With this approach, the following assertion would be valid on all platforms:

```scala
os.Path(pwd.rootPrefix) ==> pwd
```

Then `driveRoot` in this PR would become `pwd.rootPrefix`.

There is another (perhaps never used?) very subtle feature of `Windows` filesystem:  each drive has a different value for `pwd`.   It may not be necessary to model this in `os-lib`, since these values are immutable in a running jvm, and there are existing workarounds.

### Additional Details regarding unique aspects of Windows Filesystem Paths
For completeness, the following lengthy `scala` REPL session illustrates Path values returned by `java.nio.file.Paths.get()`, some of which might be surprising.
<details>

Relevant information: My system drives are C: and F:, but not J:

First, some unsurprising results:
```scala
c:\work-directory> scala.bat
scala> import java.nio.file.Paths
Welcome to Scala 3.3.1-RC5 (17.0.2, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> Paths.get("/")
val res0: java.nio.file.Path = \

scala> Paths.get("/").isAbsolute
val res2: Boolean = false

scala> Paths.get("/").toAbsolutePath
val res1: java.nio.file.Path = C:\

scala> Paths.get("c:/").toAbsolutePath
C:\work-directory

scala> Paths.get("f:/").toAbsolutePath
F:\

scala> Paths.get("f:/..").normalize.toAbsolutePath
f:\

scala> Paths.get("c:/..").normalize.toAbsolutePath
c:\
```
Now some less obvious results:
```scala
scala> Paths.get("c:").toAbsolutePath
C:\work-directory

scala> Paths.get("f:").toAbsolutePath
F:\

scala> Paths.get("f:/..").normalize.toAbsolutePath
f:\

scala> Paths.get("c:/..").normalize.toAbsolutePath
c:\

scala> Paths.get("c:..").normalize.toAbsolutePath
C:\work-directory

scala> Paths.get("f:..").normalize.toAbsolutePath
F:\..

scala> Paths.get("F:Users")
val res1: java.nio.file.Path = F:Users

scala> Paths.get("F:Users").toAbsolutePath
val res2: java.nio.file.Path = F:\work-directory\Users

scala> Paths.get("j:").toAbsolutePath
java.io.IOError: java.io.IOException: Unable to get working directory of drive 'J'
  at java.base/sun.nio.fs.WindowsPath.toAbsolutePath(WindowsPath.java:926)
  at java.base/sun.nio.fs.WindowsPath.toAbsolutePath(WindowsPath.java:42)
  ... 35 elided
Caused by: java.io.IOException: Unable to get working directory of drive 'J'
  ... 37 more
```

The error message returned for a non-existing drive  seems to imply that existing drives have a `working directory`.
This is consistent with the `Windows API` as described here:
[GetFullPathNameA](https://learn.microsoft.com/en-gb/windows/win32/api/fileapi/nf-fileapi-getfullpathnamea)

I can set the `working drive` for F: in a `CMD` session before I start up the `jvm`, and it does affect the output of `Paths.get()`.
```CMD
c:\work-directory>F:

f:\>cd work-directory

F:\work-directory>scala.bat
Welcome to Scala 3.3.1-RC5 (17.0.2, Java OpenJDK 64-Bit Server VM).
Type in expressions for evaluation. Or try :help.

scala> import java.nio.file.Paths

scala> Paths.get("/")
val res0: java.nio.file.Path = \

scala> Paths.get("/").toAbsolutePath
val res1: java.nio.file.Path = C:\

scala> Paths.get("f:")
val res2: java.nio.file.Path = f:

scala> Paths.get("f:").toAbsolutePath
val res3: java.nio.file.Path = F:\work-directory

scala> Paths.get("c:").toAbsolutePath
val res4: java.nio.file.Path = C:\work-directory

scala> Paths.get("c:")
val res5: java.nio.file.Path = c:
```

Regarding the interpretation of a drive letter expression not followed by a '/' or '\\', the
rule is that it represents the `working directory` for that drive.

These values are immutable: after the `JVM` starts up it's not possible to change a drive `working directory`.
</details>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,8 +6,13 @@ target/
 .project
 .cache
 .sbtserver
+.scala-build/
+.bsp/
 project/.sbtserver
 tags
 nohup.out
 out
 lowered.hnir
+
+# ignore vim swap files
+*.sw[op]
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,3 @@ tags
 nohup.out
 out
 lowered.hnir
-
-# ignore vim swap files
-*.sw[op]
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -400,10 +400,11 @@ object Path {
 
   def apply[T: PathConvertible](f: T, base: Path): Path = apply(FilePath(f), base)
   def apply[T: PathConvertible](f0: T): Path = {
-    val f = implicitly[PathConvertible[T]].apply(f0)
-    def nioPath = java.nio.file.Paths.get(s"$currentDrive/$f0").normalize
+    // drive letter prefix is empty unless running in Windows.
+    def nioPath = java.nio.file.Paths.get(s"$platformPrefix$f0").normalize
     val normalized = if (rootRelative(f0)) nioPath
     else {
+      val f = implicitly[PathConvertible[T]].apply(f0)
       if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount / 2) {
         throw PathError.AbsolutePathOutsideRoot
       }
@@ -440,8 +441,15 @@ object Path {
       }
     }
   }
+  /**
+   * @return true if Windows OS and path begins with slash or backslash.
+   * Examples:
+   *    rootRelative("/Users")   // true in `Windows`, false elsewhere.
+   *    rootRelative("\\Users")  // true in `Windows`, false elsewhere.
+   *    rootRelative("C:/Users") // false always
+   */
   def rootRelative[T: PathConvertible](f0: T): Boolean = {
-    if (currentWorkingDrive.isEmpty) {
+    if (platformPrefix.isEmpty) {
       false // non-Windows os
     } else {
       f0.toString.take(1) match {
@@ -450,11 +458,13 @@ object Path {
       }
     }
   }
-  def currentWorkingDrive: String = Paths.get("").toAbsolutePath.getRoot.toString match {
-    case "/" => ""
-    case s => s"$s/"
+  /**
+   * @return current working drive if Windows, empty string elsewhere.
+   */
+  val platformPrefix: String = Paths.get(".").toAbsolutePath.getRoot.toString match {
+    case "/" => ""   // implies a non-Windows platform
+    case s => s"$s/" // Windows current working drive
   }
-  def currentDrive: Path = Path(currentWorkingDrive)
 }
 
 trait ReadablePath {
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -237,10 +237,12 @@ sealed trait FilePath extends BasePath {
   def toNIO: java.nio.file.Path
   def resolveFrom(base: os.Path): os.Path
 }
+
 object FilePath {
   def apply[T: PathConvertible](f0: T) = {
-    val f = implicitly[PathConvertible[T]].apply(f0)
-    if (f.isAbsolute) Path(f0)
+    def f = implicitly[PathConvertible[T]].apply(f0)
+    // if Windows semi-absolute path, convert it to an absolute path
+    if (Path.rootRelative(f0) || f.isAbsolute) Path(f0)
     else {
       val r = RelPath(f0)
       if (r.ups == 0) r.asSubPath
@@ -297,7 +299,7 @@ object RelPath {
   def apply[T: PathConvertible](f0: T): RelPath = {
     val f = implicitly[PathConvertible[T]].apply(f0)
 
-    require(!f.isAbsolute, s"$f is not a relative path")
+    require(!f.isAbsolute && !Path.rootRelative(f0), s"$f is not a relative path")
 
     val segments = BasePath.chunkify(f.normalize())
     val (ups, rest) = segments.partition(_ == "..")
@@ -399,11 +401,14 @@ object Path {
   def apply[T: PathConvertible](f: T, base: Path): Path = apply(FilePath(f), base)
   def apply[T: PathConvertible](f0: T): Path = {
     val f = implicitly[PathConvertible[T]].apply(f0)
-    if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount / 2) {
-      throw PathError.AbsolutePathOutsideRoot
+    def nioPath = java.nio.file.Paths.get(s"$currentDrive/$f0").normalize
+    val normalized = if (rootRelative(f0)) nioPath
+    else {
+      if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount / 2) {
+        throw PathError.AbsolutePathOutsideRoot
+      }
+      f.normalize()
     }
-
-    val normalized = f.normalize()
     new Path(normalized)
   }
 
@@ -435,7 +440,21 @@ object Path {
       }
     }
   }
-
+  def rootRelative[T: PathConvertible](f0: T): Boolean = {
+    if (currentWorkingDrive.isEmpty) {
+      false // non-Windows os
+    } else {
+      f0.toString.take(1) match {
+        case "\\" | "/" => true
+        case _ => false
+      }
+    }
+  }
+  def currentWorkingDrive: String = Paths.get("").toAbsolutePath.getRoot.toString match {
+    case "/" => ""
+    case s => s"$s/"
+  }
+  def currentDrive: Path = Path(currentWorkingDrive)
 }
 
 trait ReadablePath {
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -401,10 +401,14 @@ object Path {
   def apply[T: PathConvertible](f: T, base: Path): Path = apply(FilePath(f), base)
   def apply[T: PathConvertible](f0: T): Path = {
     // drive letter prefix is empty unless running in Windows.
-    def nioPath = java.nio.file.Paths.get(s"$platformPrefix$f0").normalize
-    val normalized = if (rootRelative(f0)) nioPath
-    else {
-      val f = implicitly[PathConvertible[T]].apply(f0)
+    val normalized = {
+      val f = if (rootRelative(f0)) {
+        Paths.get(s"$platformPrefix$f0")
+      } else {
+        implicitly[PathConvertible[T]].apply(f0)
+      }
+      // in Windows, getNameCount includes the drive letter
+      // but the segment iterator doesn't include it
       if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount / 2) {
         throw PathError.AbsolutePathOutsideRoot
       }
@@ -449,11 +453,13 @@ object Path {
    *    rootRelative("\\Users")  // true in `Windows`, false elsewhere.
    *    rootRelative("C:/Users") // false always
    */
-  def rootRelative[T: PathConvertible](f0: T): Boolean = {
+  def rootRelative[T: PathConvertible](f0: T): Boolean = rootRelative(f0.toString)
+
+  def rootRelative(s: String): Boolean = {
     if (platformPrefix.isEmpty) {
       false // non-Windows os
     } else {
-      f0.toString.take(1) match {
+      s.toString.take(1) match {
         case "\\" | "/" => true
         case _ => false
       }
@@ -463,9 +469,9 @@ object Path {
   /**
    * @return current working drive if Windows, empty string elsewhere.
    */
-  val platformPrefix: String = Paths.get(".").toAbsolutePath.getRoot.toString match {
+  lazy val platformPrefix: String = Paths.get(".").toAbsolutePath.getRoot.toString match {
     case "/" => "" // implies a non-Windows platform
-    case s => s"$s/" // Windows current working drive
+    case s => s.take(2) // Windows current working drive (e.g., "C:")
   }
 }
 
@@ -483,7 +489,7 @@ class Path private[os] (val wrapped: java.nio.file.Path)
   def toSource: SeekableSource =
     new SeekableSource.ChannelSource(java.nio.file.Files.newByteChannel(wrapped))
 
-  require(wrapped.isAbsolute, s"$wrapped is not an absolute path")
+  require(wrapped.isAbsolute || Path.rootRelative(wrapped), s"$wrapped is not an absolute path")
   def segments: Iterator[String] = wrapped.iterator().asScala.map(_.toString)
   def getSegment(i: Int): String = wrapped.getName(i).toString
   def segmentCount = wrapped.getNameCount
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -242,7 +242,7 @@ object FilePath {
   def apply[T: PathConvertible](f0: T) = {
     def f = implicitly[PathConvertible[T]].apply(f0)
     // if Windows root-relative path, convert it to an absolute path
-    if (Path.rootRelative(f0) || f.isAbsolute) Path(f0)
+    if (Path.driveRelative(f0) || f.isAbsolute) Path(f0)
     else {
       val r = RelPath(f0)
       if (r.ups == 0) r.asSubPath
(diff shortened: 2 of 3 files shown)


philwalk:
I inadvertently only compiled and tested these changes with `scala 3`, and am seeing compile errors with `2.11.12`, so more work needed ...

philwalk:
Just noticed that `scalafmt` complained, will push fix.+

lefou (commented):
TBH, I have not completely made up my mind about this issue and your approach to handle it. Nevertheless, thank you for this PR. I already have some review comments to share.

The issue involves file system roots, a topic we have a dedicated issue for (#170) and also a PR (#196). We should aim for a consistent solution for both issues.

os/src/Path.scala
```diff
@@ -435,7 +441,21 @@ object Path {
...
+    if (defaultDrive.isEmpty) {
+      false // non-Windows os
+    } else {
+      f0.toString.take(1) match {
+        case "\\" | "/" => true
+        case _ => false
+      }
+    }
+  }
+  def defaultDrive: String = Paths.get("").toAbsolutePath.getRoot.toString match {
```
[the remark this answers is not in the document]
philwalk: completely agree, `currentWorkingDrive` or even `currDrive` would be much better.

philwalk:
> TBH, I have not completely made up my mind about this issue and your approach to handle it. Nevertheless, thank you for this PR. I already have some review comments to share.
> 
> The issue involves file system roots, a topic we have a dedicated issue for (#170) and also a PR (#196). We should aim for a consistent solution for both issues.

Sounds good.  I will try to see how this PR can be made compatible with (#196).  I assume we want to be compatible with Java's `Paths.get()` behavior if at all possible.

os/test/src/PathTests.scala
```diff
@@ -6,12 +6,15 @@ import os._
 import utest.{assert => _, _}
 import java.net.URI
 object PathTests extends TestSuite {
+  val wintest = Option(System.getenv("WINPATHTEST")).nonEmpty
```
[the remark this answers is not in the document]
philwalk: I added this as a convenience for experimentally enabling all tests in Windows, to see which ones fail and why.  Ideally, the platform-conditional tests can be eliminated or minimized.

philwalk:
I edited in the changes in (#196) and verified no conflicts, so the two PRs should merge without any issues.
Afterwards, I removed the other PR code and did more testing.
I applied your suggestions, made some simplifications, and squashed my commits into one.

os/src/Path.scala
```diff
@@ -435,7 +440,21 @@ object Path {
...
+        case "\\" | "/" => true
+        case _ => false
+      }
+    }
+  }
+  def currentWorkingDrive: String = Paths.get("").toAbsolutePath.getRoot.toString match {
+    case "/" => ""
+    case s => s"$s/"
+  }
+  def currentDrive: Path = Path(currentWorkingDrive)
```
lefou, line 457: Isn't this semantically the same as `root` (or `currentRoot`)?

It may sound as nitpicking, but a "drive" is a Windows-specific term (each drive resembles it own filesystem root), but we may even don't hold a drive letter at all.

lefou (commented):
Sorry for the longish review. I'm still making up my mind about this issue. It's really hard to reason about the specifics of an OS I don't use.

lefou:
Oh, and please avoid squashing while we are in the review process. It hinders to follow the incremental process. We will squash when we merge.

.gitignore
```diff
@@ -6,8 +6,13 @@ target/
...
+.scala-build/
+.bsp/
 project/.sbtserver
 tags
 nohup.out
 out
 lowered.hnir
+
+# ignore vim swap files
+*.sw[op]
```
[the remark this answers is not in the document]
philwalk: I added this after accidentally checking in the `vim` swap files, which is a bit of a nuisance to undo it in a way that doesn't show up in the git log.   I wasn't aware of `.git/info/exclude`, good suggestion.  I'll revert `.gitignore`.

os/src/Path.scala
```diff
@@ -435,7 +440,21 @@ object Path {
...
+    if (currentWorkingDrive.isEmpty) {
+      false // non-Windows os
+    } else {
+      f0.toString.take(1) match {
+        case "\\" | "/" => true
+        case _ => false
+      }
+    }
+  }
+  def currentWorkingDrive: String = Paths.get("").toAbsolutePath.getRoot.toString match {
```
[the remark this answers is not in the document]
philwalk: The two forms always return the same thing, but I agree that your suggestion is more readable.

os/test/src/PathTests.scala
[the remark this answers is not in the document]
philwalk: Some of the tests are currently disabled because expressions that are equal on other platforms differ in `Windows` by the presence or absence of a drive letter.  For example:
```scala
(root / "omg").wrapped == "C:\\omg"
Paths.get("/omg")      == "\\omg"
```
Other tests, as currently written,  are disabled on`Windows` because of mismatch between forward slashes and backslashes.
The following test would fail because "\\src\\main\\scala"  != "/src/main/scala".
```scala
val d = pwd
val abs = d / base
test("Constructor") {
  assert(
    abs.toString.drop(d.toString.length).norm == "/src/main/scala",
    abs.toString.length > d.toString.length
  )
}
```
I will revert `os/test/src/PathTests.scala` and submit a separate PR, since it's a separate concern.

os/src/Path.scala
```diff
@@ -435,7 +440,21 @@ object Path {
       }
     }
   }
-
+  def rootRelative[T: PathConvertible](f0: T): Boolean = {
```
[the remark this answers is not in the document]
philwalk: > Please add some ScalaDoc to make clear what it means and give an example of a `rootRelative` path.

ScalaDoc added.

os/src/Path.scala
philwalk: You are correct, they are the same, although using `os.root` in `os/src/Path.scala` causes circular references at compile time.

philwalk:
> Oh, and please avoid squashing while we are in the review process. It hinders to follow the incremental process. We will squash when we merge.

Makes sense.

philwalk:
After some thought, a better name than `windowsWorkingDrive` or `currentWorkingDrive` is `platformPrefix`.
The idea is that, a platform might require a prefix to resolve a `rootRelative` path.

philwalk:
I have been verifying that this PR fixes the `scala-cli` problem, and I discovered an issue.  I'm now verifying an addition change, and hope to be able to commit soon.

philwalk:
It turns out that the tests in `os/test/src/PathTests.scala` that weren't enabled when building in `Windows` revealed that a `rootRelative` path wasn't being properly checked for `PathError.AbsolutePathOutsideRoot`.

I enabled all `PathsTests.scala` tests, fixing a few of them so they pass or fail, as required.
The previously disabled tests are enabled with flag `enableTests` in place of the `Unix()` reference, to simplify seeing the changes. 

The name `rootRelative` seems misleading to me, because it is an absolute path in almost every sense.   The following expressions are equivalent except for how they are displayed.
```
Path("/omg") == Path(s"$platformPrefix/omg")
```

I just noticed an incorrect comment, I'll fix it.

philwalk:
I manually verified that this fixes the scala-cli bug https://github.com/VirtusLab/scala-cli/issues/2359

os/test/src/PathTests.scala
```diff
@@ -436,5 +439,34 @@ object PathTests extends TestSuite {
...
+      s"$platformPrefix${s.posix}"
+    }
+  }
+  implicit class ExtendOsPath(p: os.Path) {
+    def posix: String = p.toNIO.toString.posix
+    def norm: String = p.toString.norm
+  }
+  implicit class ExtendPath(p: java.nio.file.Path) {
+    def posix: String = p.toString.posix
+    def norm: String = p.toString.norm
```
lihaoyi, line 470: Let's use normal helper methods here, rather than extension methods, unless we have good reason to do otherwise.

os/test/src/PathTests.scala
```diff
@@ -3,17 +3,20 @@ package test.os
 import java.nio.file.Paths
 
 import os._
+import os.Path.{platformPrefix}
 import utest.{assert => _, _}
 import java.net.URI
 object PathTests extends TestSuite {
+  def enableTest = true
```
lihaoyi, line 10: If `enableTest` is `true`, we should be able to remove it and remove all the conditionals guarding the test cases

os/src/Path.scala
```diff
@@ -398,12 +400,18 @@ object Path {
...
+    // drive letter prefix is empty unless running in Windows.
+    val normalized = {
+      val f = if (rootRelative(f0)) {
+        Paths.get(s"$platformPrefix$f0")
+      } else {
+        implicitly[PathConvertible[T]].apply(f0)
+      }
+      if (f.iterator.asScala.count(_.startsWith("..")) > f.getNameCount / 2) {
+        throw PathError.AbsolutePathOutsideRoot
+      }
```
lihaoyi, line 412: We can re-arrange this to move the `throw` outside of the `val normalized`? All these are method-local `val`s anyway, so there's no visibility/privacy concerns, and it would be nice to remove a layer of nesting

lihaoyi:
Ieft a few comments. I am also not super familiar with windows, but if the old tests continue passing and the new tests pass then that gives us some confidence.

@philwalk could you update the PR description to:

1. Remove all the stuff about Scala-CLI. Linking to the issue in Scala-CLI should be enough for people to follow, but for a PR to OS-Lib the description of the problem should focus on exactly what's the problem with OS-Lib and not how it manifests in some downstream project

2. Explain what your solution is, in english. We already have your code in the diff, but it would be much easier to review for us and for future maintainers if the PR description had an english summary of your changes explaining what changes you needed to make, where, and why.

After that, I think I'd be OK with merging this, unless @lefou has any other concerns

os/src/Path.scala
```diff
@@ -436,6 +444,33 @@ object Path {
...
+  /**
+   * @return true if Windows OS and path begins with slash or backslash.
+   * Examples:
+   *    rootRelative("/Users")   // true in `Windows`, false elsewhere.
+   *    rootRelative("\\Users")  // true in `Windows`, false elsewhere.
+   *    rootRelative("C:/Users") // false always
+   */
+  def rootRelative[T: PathConvertible](f0: T): Boolean = rootRelative(f0.toString)
+
+  def rootRelative(s: String): Boolean = {
```
lihaoyi, line 456: Do we need this overload, given that the `implicit object StringConvertible extends PathConvertible[String]` already exists? If not we should remove it

os/test/src/PathTests.scala
philwalk: I'll remove them.

os/src/Path.scala
philwalk: Combined the two:
```scala
  def rootRelative[T: PathConvertible](f0: T): Boolean = {
    if (platformPrefix.isEmpty) {
      false // non-Windows os
    } else {
      f0.toString.take(1) match {
        case "\\" | "/" => true
        case _ => false
      }
    }
  }
```

os/src/Path.scala
philwalk: done.

os/test/src/PathTests.scala
philwalk: done (will push changes soon)

philwalk:
> After that, I think I'd be OK with merging this, unless @lefou has any other concerns
All your suggestions are implemented, thanks for the feedback!

philwalk:
I merged #196 with this PR as an experiment, and all tests pass, although I think there may be a simplification to this PR that is now possible.

os/src/Path.scala
```diff
@@ -436,6 +441,32 @@ object Path {
...
+        case _ => false
+      }
+    }
+  }
+
+  /**
+   * @return current working drive if Windows, empty string elsewhere.
+   * Paths.get(platformPrefix) == current working directory on all platforms.
+   */
+  lazy val platformPrefix: String = Paths.get(".").toAbsolutePath.getRoot.toString match {
```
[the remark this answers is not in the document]
philwalk: How about `drivePrefix`?

PR #210, opened 2023-09-15
os.proc.call: replace reference to Chunk(/ByteData).string() with text()

Flowdalic:
The .string() method in geny's ByteData was removed [1].

1: https://github.com/com-lihaoyi/geny/commit/d1289bc020e79df9482c2cb46c65a1c41511c24b


lefou (approved):
Thanks!

PR #211, opened 2023-09-18
Add external mappings for scaladoc

Flowdalic:
External mappings allow linking to other project's documentation, which makes it easier for users to understand the API.

This is a WIP. The main thing to address is that the place where we set scalaDocOptions may be not correct, I could imagine that it should be done in `OsJvmModule`. Furthermore, I should be only set if `scaladoc` from Scala 3 is used, as the `-external`mappings` option is only available from there on. Unfortunately, I lack the mill and os-lib build system knowledge to proceed here. Any feedback would be appreciated.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -89,6 +89,23 @@ trait OsModule extends OsLibModule { outer =>
   def ivyDeps = Agg(Deps.geny)
 
   def artifactName = "os-lib"
+
+  val scalaDocExternalMappings = Seq(
+    ".*scala.*::scaladoc3::https://scala-lang.org/api/3.x/",
+    ".*java.*::javadoc::https://docs.oracle.com/javase/8/docs/api/",
+    s".*geny.*::scaladoc3::https://javadoc.io/doc/com.lihaoyi/geny_3/${Deps.geny.dep.version}/",
+  ).mkString(",")
+
+  def conditionalScalaDocOptions: T[Seq[String]] = T {
+    if (ZincWorkerUtil.isDottyOrScala3(scalaVersion()))
+      Seq(
+        s"-external-mappings:${scalaDocExternalMappings}"
+      )
+    else Seq()
+  }
+
+  def scalaDocOptions = super.scalaDocOptions() ++ conditionalScalaDocOptions()
+
 }
 
 object os extends Module {


Flowdalic:
Updated change to include `(ZincWorkerUtil.isDottyOrScala3(scalaVersion())` condition check. Thanks @lefou

lefou (approved):
Looks good to me. Is there no such option for Scala 2, or why do you just link when we build for Scala 3?

SethTisue:
Scala 2 Scaladoc definitely supports source links; see for example the "Source: ..." link at https://www.scala-lang.org/api/2.13.11/scala/collection/immutable/List.html   . it's via `-doc-source-url`

lefou:
> Scala 2 Scaladoc definitely supports source links; see for example the "Source: ..." link at https://www.scala-lang.org/api/2.13.11/scala/collection/immutable/List.html . it's via `-doc-source-url`
> 
> be careful about [scala/bug#12867](https://github.com/scala/bug/issues/12867) though — a behavior changed between 2.13.11 and 2.13.12

This PR isn't so much about source links but about cross-references to scaladoc of API dependencies.

SethTisue:
Oh, sorry, I read too quickly.

But yes, Scala 2 Scaladoc has that too, via `-doc-external-doc`

SethTisue:
> Even `cs install scaladoc:2.13.12` isn't working 

Worked for me, just now, so maybe that could use its own ticket.

> Is there any documentation about it? 

It appears there isn't. https://docs.scala-lang.org/overviews/scaladoc/for-library-authors.html is where it would be if it existed.

`-doc-external-doc` (formerly `-doc-external-uris`) is so old it goes back to when Scala was hardly more than a glimmer in Martin's eye, and nobody has documented it since afaict. GitHub search has a number of examples, though: https://github.com/search?q=-doc-external-doc&type=code . And https://github.com/ThoughtWorksInc/sbt-api-mappings is probably a source of knowhow, too.

lefou:
I played a bit (probably an understatement) with `-doc-external-doc` and `-jdk-api-doc-base`, but both produced links to some URLs, but not the correct ones. I think we should just go with reference link in Scala 3 and be happy it works there reasonable. Thanks @SethTisue for you input.

lefou (approved):

lefou:
Thank you, @Flowdalic !

PR #236, opened 2023-11-17
Watch windows

sake92:
There are 2 parts to this fix:
1. the `FILE_TREE` modifier, which tells Windows to watch the whole folder recursively. This prevents errors when trying to move/rename a folder, since otherwise it locks them and we get errors.. Got this from here https://stackoverflow.com/a/6265860/4496364
2. preventing dynamic registration of subfolders in the `recursiveWatches` method

The diff in the test is a bit messed up, there are just a few lines changed.. :/


lefou (approved):
Looks good to me! Thank you!

PR #238, opened 2023-11-18
Fix propagateEnv = false to not propagate env

lolgab:
Originated from https://github.com/com-lihaoyi/mill/issues/2881

Pull request: https://github.com/com-lihaoyi/os-lib/pull/238

### DIFF
--- a/os/src-jvm/ProcessOps.scala
+++ b/os/src-jvm/ProcessOps.scala
@@ -337,12 +337,17 @@ private[os] object ProcessOps {
   ): ProcessBuilder = {
     val builder = new java.lang.ProcessBuilder()
 
-    val baseEnv =
-      if (propagateEnv) sys.env
-      else Map()
-    for ((k, v) <- baseEnv ++ Option(env).getOrElse(Map())) {
-      if (v != null) builder.environment().put(k, v)
-      else builder.environment().remove(k)
+    val environment = builder.environment()
+
+    if (!propagateEnv) {
+      environment.clear()
+    }
+
+    if (env != null) {
+      for ((k, v) <- env) {
+        if (v != null) builder.environment().put(k, v)
+        else builder.environment().remove(k)
+      }
     }
 
     builder.directory(Option(cwd).getOrElse(os.pwd).toIO)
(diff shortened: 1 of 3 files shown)


os/src-jvm/ProcessOps.scala
```diff
@@ -337,12 +337,17 @@ private[os] object ProcessOps {
...
-    val baseEnv =
-      if (propagateEnv) sys.env
-      else Map()
-    for ((k, v) <- baseEnv ++ Option(env).getOrElse(Map())) {
-      if (v != null) builder.environment().put(k, v)
-      else builder.environment().remove(k)
+    val environment = builder.environment()
+
+    if (!propagateEnv) {
+      environment.clear()
```
lolgab, line 343: `java.lang.ProcessBuilder` propagates the environment by default. If you don't want to propagate it you need to `clear()` the map.

os/test/src-jvm/SubprocessTests.scala
```diff
@@ -104,17 +104,41 @@ object SubprocessTests extends TestSuite {
...
-        assert(res1.out.lines() == Seq("Hello12"))
+        locally {
+          val res3 = proc("bash", "-c", "echo 'Hello'$ENV_ARG").call(env = Map("ENV_ARG" -> "123"))
+          assert(res3.out.lines() == Seq("Hello123"))
+        }
 
-        val res2 = proc("bash", "-c", "echo 'Hello$ENV_ARG'").call(env = Map("ENV_ARG" -> "12"))
-        assert(res2.out.lines() == Seq("Hello$ENV_ARG"))
+        locally {
+          val res4 = proc("bash", "-c", "echo \"$TEST_SUBPROCESS_ENV\"").call(
```
lefou, line 128: Can you add a comment, that we expect this env var to be set from the build tool / test runner?

lefou (commented):
Nice catch! Looks good to me. One nit below.

os/src-jvm/ProcessOps.scala
lefou: Or even assert that it is set as expected.

os/test/src-jvm/SubprocessTests.scala
lefou: Or even assert that it is set as expected.

lefou (approved):

Issue #276, opened 2024-06-11
`os.root.baseName` throws an exception


philwalk:
A new `scala-cli` bug is caused by `os.root.baseName` ~being an empty string~ throwing an exception.

https://github.com/VirtusLab/scala-cli/issues/2954
If a `scala-cli` script resides in the root directory, it crashes with an error message:
```scala
Error: os.PathError$LastOnEmptyPath: empty path has no last segment
```
Although this could be avoided in client code, it would require any `os.Path` object to be tested before referencing `baseName`.

One way to avoid the exception would be to redefine `os.root.baseName` as the empty String.   However, the `jvm` defines the empty String as `pwd`, so if `os.root.baseName` is also the empty String, that would make it equal to `pwd`, which is a problem.

Shell environments (mostly) treat the root directory as its own parent, which is much easier to work with.  For example, `/usr/bin/basename` takes this approach.   The expression `basename /` resolves to `/`.

Defining `os.root.baseName` as `s"$driveRoot/"` resolves the `scala-cli` bug.

This is the approach taken by PR #277

philwalk:
Clarification: the original symptom for the `scala-cli` bug was a `LastOnEmptyPath` error.

PR #277, opened 2024-06-11
make os.root.baseName equal to s"${os.root/}" rather than throwing an exception

philwalk:
This addresses the problems described by #276
It provides a fix for https://github.com/VirtusLab/scala-cli/issues/2954
It makes `os.root.last` the empty string rather than throwing `LastOnEmptyPath`.
Then, it defines `os.root.baseName`  as `s"$driveRoot/"`, rather than throwing an Exception.

### DIFF
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -202,12 +202,16 @@ trait BasePathImpl extends BasePath {
   }
 
   override def baseName: String = {
-    val li = last.lastIndexOf('.')
-    if (li == -1) last
-    else last.slice(0, li)
+    lastOpt match {
+      case None => s"${Path.driveRoot}/"
+      case Some(lastSegment) =>
+        val li = lastSegment.lastIndexOf('.')
+        if (li == -1) last
+        else lastSegment.slice(0, li)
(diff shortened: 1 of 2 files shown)


lihaoyi:
TBH I'm not sure if this is the correct behavior for this. I think it's reasonable for it to fail and downstream to handle it, v.s. arbitrarily returning empty strings. 

There are tons of cases where empty strings are not what you want either, and they have a tendency to propagate far beyond the point of creation before misbehaving in complicated ways. In comparison, the exception immediately tells you what's going on and you can remediate it at the failure site

philwalk:
> TBH I'm not sure if this is the correct behavior for this. I think it's reasonable for it to fail and downstream to handle it, v.s. arbitrarily returning empty strings.

I don't disagree with your reasoning.   A concern is that throwing of an exception makes it unsafe to call `.baseName` on any `os.Path` that might be === `os.root`.   Empty string is a reasonable way to represent that a path has no segments, implying that it's `os.root`.   This is similar to the way that `dirname /` returns `/` rather than an error message.

Issue #282, opened 2024-06-26
Automatically pump streams with `os.Inherit` on changed `System.out` `System.err` `System.in`


lolgab:
In Mill server-client mode the standard streams are changed using `System.setOut`, etc. methods so they can be propagated from the server to the client and shown to the user's console.
Unfortunately, when running processes with `os.proc` and `os.Inherit` the changes done with `System.setOut`, etc. are not honored, since the `java.lang.ProcessBuilder` doesn't honor the changes in System streams.
This issue is about natively support honoring the changes in streams in `os.proc` without users having to manually redirect the output and the input via pipes.
Since there is no way to know if the streams were changed, os-lib would need to add an API to allow users to set the original streams, like:
```scala
os.proc.setOriginalSystemOut(System.out)
os.proc.setOriginalSystemErr(System.err)
os.proc.setOriginalSystemIn(System.in)

// or
os.proc.setOriginalStreams()
// which would do the same things but on the background
```
So when a process is executed and `System.in ne os.proc.originalSystemIn && stdin == os.Inherit` it would pump the data to the `System.in` using `os.Pipe` (as we currently do in Mill), since it assumes it is not the original `System.in`.
The same thing would happen with `System.out` and  `System.err`.

To work around this problem, Mill implemented some helper methods, but they are bespoke and are not what users would want to write.
In fact, many plugins and users' builds in the wild just use `os.proc` with `os.Inherit` since it's the easiest way, and then they see broken output in server-client mode.

PR #283, opened 2024-07-19
Add API to redirect `os.Inherit`ed streams globally to be consistent with std streams redirections

lihaoyi:
Fixes https://github.com/com-lihaoyi/os-lib/issues/282

Unlike usage of `System.in`/`System.out`/`System.err` programmatically, or the `scala.Console` equivalents, output from subprocesses with `os.Inherit` is not affected by `System.setIn`/`setOut`/`setErr`. This is often counterintuitive, because you expect redirecting these streams to redirect all output, only to find output from subprocess leaking through a side channel to the original process `in`/`out`/`err`.

This PR adds the `os.Inherit.in`/`out`/`err` dynamic variables, which can be used to re-assign `os.Inherit` on a scoped threadlocal basis. This allows developers who use `System.setOut` or `Console.withOut` to also redirect subprocess output within the scope of the stdout redirect, without users working within those scopes to have to remember to redirect them manually at every callsite.

Because we do not control `System.setOut` or `Console.withOut`, there isn't really a way for us to detect and do this automatically (e.g. redirects may happen even before OS-Lib has been even classloaded) and so we have to rely on it being opt-in.

As an escape hatch, in case users do not want this behavior, we provide a `os.Inherit0` redirect which performs identically to the original `os.Inherit` even in the process of redirecting `os.Inherit.{in,out,err}`

Covered by an added unit test

### DIFF
--- a/Readme.adoc
+++ b/Readme.adoc
@@ -2224,6 +2224,11 @@ string, int or set representations of the `os.PermSet` via:
 
 == Changelog
 
+[#0-10-3]
+=== 0.10.3
+
+* `os.Inherit`
+
 [#0-10-2]
 === 0.10.2
 
--- a/Readme.adoc
+++ b/Readme.adoc
@@ -2224,8 +2224,8 @@ string, int or set representations of the `os.PermSet` via:
 
 == Changelog
 
-[#0-10-3]
-=== 0.10.3
+[#main]
+=== main
 
 * `os.Inherit`
 
--- a/os/src/ProcessOps.scala
+++ b/os/src/ProcessOps.scala
@@ -119,6 +119,18 @@ case class proc(command: Shellable*) {
 
     val cmdChunks = commandChunks
     val commandStr = cmdChunks.mkString(" ")
+    val resolvedStdin = stdin match {
+      case os.Inherit => os.Inherit.in
+      case v => v
+    }
+    val resolvedStdout = stdout match {
+      case os.Inherit => os.Inherit.out
+      case v => v
+    }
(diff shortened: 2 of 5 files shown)


lefou (commented):
Looks reasonable. Can you add it to the documentation, too?

lihaoyi:
I added a line in the changelog; we can postpone adding it to the main readme body until after it is published to avoid confusion

lefou:
IMHO, we shouldn't postpone the documentation of a new feature to a subsequent release. It would be missing in the tagged release then and not available for offline reading, for example.

We could add a `since <next version>` marker to the new doc section, which would be helpful in any case.

lihaoyi:
Updating it together with tagging/publishing the release would ensure the smallest gap between what's documented and what's published. I don't want people to read docs and have to consciously skip parts that dont apply

It's a compromise, but some compromise is inevitable when most people will see the docs on the web from the latest version of `main` while releases are done less frequently.

lefou:
It's your call.

----

I kind of disagree, though. It's what we did in the past and what I and others do in other projects. The Readme and other documentation should reflect the current state of the branch it belongs to. Looking at some possibly unreleased state is an inherent property of looking at the `main` branch. In contrast to looking at a tag or release branch. It's a good thing to keep features and their docs close, ideally in the same PR. That enables use to cut a release including its documentation at any point in time.

Users can look at a specific tag rather easily. The docs then always reflects that version.

To provide stable documentation reflecting a specific version, we could and probably should just copy a rendered version to `ghpages`. That can be automated.

PR #286, opened 2024-07-26
Added proper SIGTERM/SIGKILL handling for sub-processes

### LINKED ISSUE #284: `os.proc.call`'s `timeout` has a termination race-condition from `SIGTERM` and `SIGKILL`
The current implementation of `os.proc.call`'s `timeout` flag uses two consecutive calls to `p.destroy()` and `p.forciblyDestroy()`. The effects of these calls are to first send `SIGTERM`, and then send `SIGKILL` to the process.

The roles of these two signals are:
* `SIGTERM`: instruct the process to terminate, the process may intercept this and perform necessary clean-up operations, or may decide to ignore it entirely
* `SIGKILL`: instruct the process to terminate immediately -- this signal cannot be intercepted.

By sending these two signals back-to-back, the parent process produces a race-condition between how quickly the child can execute its `SIGTERM` handler and clean up resources and the issue of the `SIGKILL`. In my experiments, I've found that `SIGKILL` it the cause of process exit the vast majority of the time. This means that the (potentially necessary) clean-up of the process is often not performed or worse interrupted. If the handler itself contained code to write file contents back to disk, modify a database, and so on, these operations may be corrupted. If the child process itself has children that need terminating, this could not be issued, leading to the parent process hanging. 

## What are the possible desired outcomes?
There are three ways that the timeout should be terminating the process:

1. Only send `SIGTERM`: it doesn't matter how long it takes, we need to ensure safe clean-up
2. Only send `SIGKILL`: the process has no important state, it should be terminated immediately
3. Send `SIGTERM`, wait an appropriate amount of time, then send `SIGKILL`: we want to offer the process an opportunity to clean-up, but if this takes too long (perhaps the clean-up process itself is hanging), we want to forcibly terminate -- this is the scenario done by `os.lib`, albeit without allowing sufficient time to perform the handler.

## What is normally done?
The `SIGKILL` signal is useful to issue when a process is not responding in a timely fashion to its `SIGTERM` event and the two are usually sent together with a _delay_. The Linux `timeout` command offers this with the `-k n` flag, which sends a `SIGKILL`  signal `n` seconds after the original timeout sent `SIGTERM`.

## Solutions
The race condition caused by the consecutive calls to `destroy` and `forciblyDestroy` is a bug, and could be addressed by supporting a similar system allowing outcomes (1), (2), or (3) configurably and safely.

For backwards compatibility, however, it might be wise to just support (1) with the current system.

j-mie6:
Fixes #284.

The idea is to introduce a new flag to both `os.proc.call` and `os.ProcGroup.call`, as well as backing implementations in `os.ProcessLike`. The flag, `timeoutGracePeriod`, allows for a configurable use of `SIGKILL`, which forcibly kills a process, after the issue of a `SIGTERM`, which allows a process to clean up. The values for the flag have the following behaviours:

|value | behaviour|
|------|-----------|
|  -1        | When `timeout != -1`, only a `SIGTERM` will be issued, requiring the child process to gracefully terminate.  |
|  0        | When `timeout != -1`, only a `SIGKILL` is issued, demanding the child terminate immediately with no chance of graceful cleanup.               |
|  n > 0  |  `timeout != -1`, first issue a `SIGTERM` and then wait for a further `n` milliseconds before issuing a `SIGKILL`, this provides a reasonable timeframe for process cleanup, but accounts for misbehaving processes.              |

For now, the default has been set to `timeoutGracePeriod = 1000`.


os/src/SubProcess.scala
```diff
@@ -120,7 +126,8 @@ class SubProcess(
    * limit is given, explicitly destroys the subprocess if it has not completed by
    * the time the timeout has occurred
    */
-  def join(timeout: Long = -1): Boolean = {
+  def join(timeout: Long = -1, timeoutGracePeriod: Long = 1000): Boolean = {
+    // FIXME:
     val exitedCleanly = waitFor(timeout)
```
j-mie6, line 131: The current logic for `join` (undocumented) returns `true` if the process had already terminated by the time the timeout completes. If the process is required to terminate in anyway, this will be `false`.

Would it be better instead to have `exitedCleanly` also mean "exited within the grace period and accepted `SIGTERM` without needing `SIGKILL`"? I would say that also feels like a "clean exit", in that no forcible termination was required.

os/src/SubProcess.scala
```diff
@@ -339,28 +353,40 @@ class ProcessPipeline(
...
+        else {
+          destroy()
+          if (!waitFor(timeoutGracePeriod)) {
+            destroyForcibly()
+          }
+        }
+        waitFor(-1)
+        // note that this is the only part that isn't shared with the other implementation of join... they could be unified if this is made an
+        // abstract method: then this implementation can move to the superclass (except for the default -1 case above, but that could be done via an override)
+        processes.flatMap(_.outputPumperThread).foreach(_.join())
```
j-mie6, line 386: These are the only bit of this implementation that isn't shared with the single process version. If `joinPumperThreads` was created as a new `protected` abstract method on the super-trait, then both implementations of `join` can be cleanly factored out as a default implementation in the trait itself, reducing code repetition.

Thoughts?

j-mie6:
@lihaoyi other than the documentation for `ProcessLike`, and any additional tests, this seems pretty much good to go. I have a couple of questions above that it would be nice to discuss. Can I get a CI run please?

lihaoyi:
I think this looks good. 

1000 seems a bit long for a grace period; what do you think about something shorter, like 50ms or 100ms? I don't want to change the behavior too much from the status quo, which terminates the process immediately

os/src/SubProcess.scala
```diff
@@ -49,19 +49,27 @@ sealed trait ProcessLike extends java.lang.AutoCloseable {
...
+  // FIXME: docs
   /**
    * Wait up to `millis` for the [[ProcessLike]] to terminate and all stdout and stderr
    * from the subprocess to be handled. By default waits indefinitely; if a time
    * limit is given, explicitly destroys the [[ProcessLike]] if it has not completed by
    * the time the timeout has occurred
    */
-  def join(timeout: Long = -1): Boolean
+  def join(timeout: Long = -1, timeoutGracePeriod: Long = 1000): Boolean
+
```
j-mie6, line 60: The current logic for `join` (undocumented) returns `true` if the process had already terminated by the time the timeout completes. If the process is required to terminate in anyway, this will be `false`.

Would it be better instead to have `exitedCleanly` also mean "exited within the grace period and accepted `SIGTERM` without needing `SIGKILL`"? I would say that also feels like a "clean exit", in that no forcible termination was required.

j-mie6:
> I think this looks good.
> 
> 1000 seems a bit long for a grace period; what do you think about something shorter, like 50ms or 100ms? I don't want to change the behavior too much from the status quo, which terminates the process immediately

Sure, 100ms is fine. Do note that it's unlikely to wait the full 1000ms anyway, because SIGTERM exits fast unless there is cleanup to do or the process ignores it anyway.

os/src/SubProcess.scala
lihaoyi: I think the current behavior is probably fine; especially for backwards compat we should leave it in place. If people want more control over exactly what happens in the scenario where things exit during the grace period, they can always wrap os.proc.spawn` themselves

j-mie6: Ok! I'll make sure that this behaviour is clearly documented then

lihaoyi:
I think this looks good. @j-mie6 if you're happy with too it I'll merge it and cut a release

j-mie6:
I've patched up the documentation now, so I'm happy to ship it if you are :)

PR #295, opened 2024-08-29
Introduce `os.SubProcess.env` `DynamicVariable` to override default `env`

lolgab:
In Mill and any client-server application, it is handy to start processes on the server passing the client environment. This requires to pass the environment manually in all the `os.proc(...).call()`s. An easier alternative is to override the environment as we to for `os.Inherit` `out`, `in` and `err` pipes. So users can forget about the environment and have the right one passed automatically.

Pull request: https://github.com/com-lihaoyi/os-lib/pull/295

### DIFF
--- a/Readme.adoc
+++ b/Readme.adoc
@@ -1676,7 +1676,7 @@ It is possible to customize the default environment passed to subprocesses by se
 ----
 val clientEnvironment: Map[String, String] = ???
 os.SubProcess.env.withValue(clientEnvironment) {
-  os.proc(command).call() // clientEnvironment is passed by default instead of the system environment
+  os.call(command) // clientEnvironment is passed by default instead of the system environment
 }
 ----
 
--- a/Readme.adoc
+++ b/Readme.adoc
@@ -1667,6 +1667,19 @@ val sha = os.spawn(cmd = ("shasum", "-a", "256"), stdin = gzip.stdout)
 sha.stdout.trim ==> "acc142175fa520a1cb2be5b97cbbe9bea092e8bba3fe2e95afa645615908229e  -"
 ----
 
+==== Customizing the default environment
+
+Client-server CLI applications sometimes want to run subprocesses on the server based on the environment of the client.
+It is possible to customize the default environment passed to subprocesses by setting the `os.SubProcess.env` threadlocal:
+
+[source,scala]
+----
+val clientEnvironment: Map[String, String] = ???
+os.SubProcess.env.withValue(clientEnvironment) {
+  os.proc(command).call() // clientEnvironment is passed by default instead of the system environment
+}
+----
+
 == Spawning Pipelines of Subprocesses
 
 After constructing a subprocess with `os.proc`, you can use the `pipeTo` method
--- a/build.sc
+++ b/build.sc
@@ -51,7 +51,8 @@ trait SafeDeps extends ScalaModule {
 }
 
 trait MiMaChecks extends Mima {
-  def mimaPreviousVersions = Seq("0.9.0", "0.9.1", "0.9.2", "0.9.3", "0.10.0", "0.10.1", "0.10.2", "0.10.3", "0.10.4")
+  def mimaPreviousVersions =
+    Seq("0.9.0", "0.9.1", "0.9.2", "0.9.3", "0.10.0", "0.10.1", "0.10.2", "0.10.3", "0.10.4")
   override def mimaBinaryIssueFilters: T[Seq[ProblemFilter]] = Seq(
     ProblemFilter.exclude[ReversedMissingMethodProblem]("os.PathConvertible.isCustomFs"),
     // this is fine, because ProcessLike is sealed (and its subclasses should be final)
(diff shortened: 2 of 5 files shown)


lolgab:
This should allow us to fix https://github.com/com-lihaoyi/mill/issues/2881

os/src/ProcessOps.scala
```diff
@@ -251,6 +251,19 @@ case class proc(command: Shellable*) {
...
+
+  /**
+   * The env passed by default to child processes
+   */
+  val env = new scala.util.DynamicVariable[Map[String, String]](sys.env)
+
+  // TODO: Delete when in binary compatibity breaking window
+  def andThen[T](f: proc => T): Seq[Shellable] => T = s => f(apply(s))
+  // TODO: Delete when in binary compatibity breaking window
+  def compose[T](f: T => Seq[Shellable]): T => proc = t => apply(f(t))
```
lefou, line 264: Is this related to this PR? Why is it here?

build.sc
```diff
@@ -51,11 +51,12 @@ trait SafeDeps extends ScalaModule {
...
 
 trait MiMaChecks extends Mima {
-  def mimaPreviousVersions = Seq("0.9.0", "0.9.1", "0.9.2", "0.9.3", "0.10.0")
+  def mimaPreviousVersions = Seq("0.9.0", "0.9.1", "0.9.2", "0.9.3", "0.10.0", "0.10.1", "0.10.2", "0.10.3", "0.10.4")
   override def mimaBinaryIssueFilters: T[Seq[ProblemFilter]] = Seq(
     ProblemFilter.exclude[ReversedMissingMethodProblem]("os.PathConvertible.isCustomFs"),
     // this is fine, because ProcessLike is sealed (and its subclasses should be final)
-    ProblemFilter.exclude[ReversedMissingMethodProblem]("os.ProcessLike.joinPumperThreadsHook")
+    ProblemFilter.exclude[ReversedMissingMethodProblem]("os.ProcessLike.joinPumperThreadsHook"),
+    ProblemFilter.exclude[MissingTypesProblem]("os.proc$")
```
lolgab, line 59: I am not sure if this is a real problem.
I tried in many ways to add this but I couldn't.
The way the companion object is generated when you have a case class with a vararg argument seems not reproducible, by writing the type by hand. For example: what is the type of a Vararg? 🤔 `Seq[Shellable]` while correct for binary compatibility is not the same at compile time.

os/src/ProcessOps.scala
lolgab: Scala 2 automatically adds `extends AbstractFunctionN` to all case classes with less than 6 (IIRC) arguments.
Here I couldn't find a way to add it since I don't know the type of a Vararg in Scala, or if it even exists. But these methods were created synthetically by Scala 2 and when you add a manual companion object they are missing.
I wanted to create them only for Scala 2 but I think it's fine to added them to Scala 3 as well, since they will be gone when we release os-lib `0.11.0`

build.sc
lefou: Could you post the output of MiMa?

lolgab:
```
Found 3 issue when checking against com.lihaoyi:os-lib_2.12:0.10.4
 * static method andThen(scala.Function1)scala.Function1 in class os.proc does not have a correspondent in current version
   filter with: ProblemFilter.exclude[DirectMissingMethodProblem]("os.proc.andThen")
 * static method compose(scala.Function1)scala.Function1 in class os.proc does not have a correspondent in current version
   filter with: ProblemFilter.exclude[DirectMissingMethodProblem]("os.proc.compose")
 * the type hierarchy of object os.proc is different in current version. Missing types {scala.runtime.AbstractFunction1}
   filter with: ProblemFilter.exclude[MissingTypesProblem]("os.proc$")
```

lihaoyi: e.g. we can put it on os.Subprocess instead, since now we os.call and os.spawn not everything goes through os.proc anymore

lefou (commented):
Looks good to me. We need to mention the concept in the readme/docs and give an example.

os/src/ProcessOps.scala
lolgab: Not a problem anymore after the change to `os.SubProcess.env`

lefou (approved):

Readme.adoc
```diff
@@ -1667,6 +1667,19 @@ val sha = os.spawn(cmd = ("shasum", "-a", "256"), stdin = gzip.stdout)
...
+==== Customizing the default environment
+
+Client-server CLI applications sometimes want to run subprocesses on the server based on the environment of the client.
+It is possible to customize the default environment passed to subprocesses by setting the `os.SubProcess.env` threadlocal:
+
+[source,scala]
+----
+val clientEnvironment: Map[String, String] = ???
+os.SubProcess.env.withValue(clientEnvironment) {
+  os.proc(command).call() // clientEnvironment is passed by default instead of the system environment
```
lefou, line 1679:
```suggestion
  os.call(command) // clientEnvironment is passed by default instead of the system environment
```

PR #297, opened 2024-09-05
Allow multiple segments in Stringliterals

pawelsadlo:
This PR adds support for multi-segment literal strings.
for example
```scala
val path = root / "foo/bar" 
val qux = "qux"
val path = root / "foo/bar" / "baz" / qux
val path = root / "foo/bar" / "baz/qux"
```
it also parses `..` segments from literal as `os.up` enabling syntax like:
```scala
val path = root / "foo" / ".." / "bar" // equivalent to `root / "foo" / os.up / "bar"`
val path = root / "foo" /  "../bar" // equivalent to `root / "foo" / os.up / "bar"`
```
non-canonical paths used in literals result in compile-time errors, suggesting usage of canonical paths or removing path segment, eg.
```scala
val path = root / "foo/./bar" //suggests "foo/bar"
val path = root / "foo//bar" //suggests "foo/bar"
val path = root / "//foo//bar/./baz" //suggests "foo/bar/baz"

val path = root / ""  //suggests removing the segment
val path = root / "." //suggests removing the segment
val path = root / "/" //suggests removing the segment
val path = root / "//" //suggests removing the segment
val path = root / "/./" //suggests removing the segment

```

Note:
Its not usable in os-Lib itself, due to the fact that it would lead to macro expansion in the same compilation unit as its definition.

@lihaoyi 
there is a little bit of hacking involved:

1. There is a default implicit conversion not being a macro to be used within os library, without this we would have to add a submodule and split the whole project, I'm not even sure if its doable.
4. Needed to turn off acyclic in `Path` and particular `Macro` files (also needed to mock `acyclic.skipped` in case of `scala 3`).
5. Needed to provide another implicit conversion in `ViewBoundImplicit` trait because macros turn out to be not avaliable as implicit views, this is needed for `ArrayPathChunk` and `SeqPathChunk` to work.

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -24,6 +24,7 @@ object Deps {
   val geny = ivy"com.lihaoyi::geny::1.1.1"
   val sourcecode = ivy"com.lihaoyi::sourcecode::0.4.2"
   val utest = ivy"com.lihaoyi::utest::0.8.4"
+  def scalaReflect(scalaVersion: String) = ivy"org.scala-lang:scala-reflect:$scalaVersion"
   def scalaLibrary(version: String) = ivy"org.scala-lang:scala-library:${version}"
 }
 
@@ -94,6 +95,13 @@ trait OsLibModule
 
 trait OsModule extends OsLibModule { outer =>
   def ivyDeps = Agg(Deps.geny)
+  override def compileIvyDeps = T{
+    scalaVersion.zip(super.compileIvyDeps).map{
+      case (scalaVer,superDeps) =>
+        val scalaReflectOpt = if (!ZincWorkerUtil.isDottyOrScala3(scalaVer)) Some(Deps.scalaReflect(scalaVer)) else None
+        superDeps ++ Agg.from(scalaReflectOpt)
+    }
+  }
 
   def artifactName = "os-lib"
 
--- a/build.sc
+++ b/build.sc
@@ -96,11 +96,10 @@ trait OsLibModule
 trait OsModule extends OsLibModule { outer =>
   def ivyDeps = Agg(Deps.geny)
   override def compileIvyDeps = T{
-    scalaVersion.zip(super.compileIvyDeps).map{
-      case (scalaVer,superDeps) =>
-        val scalaReflectOpt = if (!ZincWorkerUtil.isDottyOrScala3(scalaVer)) Some(Deps.scalaReflect(scalaVer)) else None
-        superDeps ++ Agg.from(scalaReflectOpt)
-    }
+    val scalaReflectOpt = Option.when(!ZincWorkerUtil.isDottyOrScala3(scalaVersion())) (
+      Deps.scalaReflect(scalaVersion())
+    )
+    super.compileIvyDeps() ++ scalaReflectOpt
   }
 
   def artifactName = "os-lib"
--- a/os/src/Path.scala
+++ b/os/src/Path.scala
@@ -2,10 +2,12 @@ package os
 
 import java.net.URI
 import java.nio.file.Paths
-
 import collection.JavaConverters._
 import scala.language.implicitConversions
-import acyclic.skipped //needed for cross-version defined macros
+import acyclic.skipped
+import os.PathError.{InvalidSegment, NonCanonicalLiteral}
+
+import scala.util.Try //needed for cross-version defined macros
 
 trait PathChunk {
   def segments: Seq[String]
@@ -23,6 +25,25 @@ object PathChunk extends PathChunkMacros {
     val strNoTrailingSeps = s.dropRight(trailingSeparatorsCount)
     val splitted = strNoTrailingSeps.split('/')
     splitted ++ Array.fill(trailingSeparatorsCount)("")
+  }
+
+  private[os] def segmentsFromStringLiteralValidation(literal: String) = {
+    val stringSegments = segmentsFromString(literal)
+    val validSegmnts = validLiteralSegments(stringSegments)
+    val sanitizedLiteral = validSegmnts.mkString("/")
+    if (validSegmnts.isEmpty) throw InvalidSegment(
+      literal,
+      s"Literal path sequence [$literal] doesn't affect path being formed, please remove it"
+    )
+    if (literal != sanitizedLiteral) throw NonCanonicalLiteral(literal, sanitizedLiteral)
+    stringSegments
+  }
+  private def validLiteralSegments(segments: Array[String]): Array[String] = {
+    val AllowedLiteralSegment = ".."
+    segments.collect {
+      case AllowedLiteralSegment => AllowedLiteralSegment
+      case segment if Try(BasePath.checkSegment(segment)).isSuccess => segment
+    }
   }
 
   implicit class RelPathChunk(r: RelPath) extends PathChunk {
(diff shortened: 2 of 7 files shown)


pawelsadlo:
@lihaoyi could you rerun checks?

pawelsadlo:
@lihaoyi 
could you take a look at checks after [924a02e]? 
did you encounter something like this before?

`adding more tests[5a49ac2]` succeed in (ubuntu-latest, 17) and almost nothing changed since then (i have only added binary compatibility shims)

After last failed workflow I have changed macro to make StringPathChunk and ArrayPathChunk instead of SubPathChunk, but I dont know why SubPathChunk would fail..

lihaoyi:
@pawelsadlo will take a look!

pawelsadlo:
@lihaoyi it succeed now, so I dont know if we need to examine it, anyway could do a review?

os/src/Path.scala
```diff
@@ -5,29 +5,52 @@ import java.nio.file.Paths
...
 }
-object PathChunk {
+trait StringPathChunkConversion {
+
+  implicit def stringToPathChunk(s: String): PathChunk =
+    new PathChunk.StringPathChunkInternal(s)
+}
+
+object PathChunk extends PathChunkMacros {
+  def segmentsFromString(s: String): Array[String] = {
```
pawelsadlo, line 21: @lihaoyi this is public only to make tests possible, however if you are fine with deleting tests for it, we can reduce accessibility

lihaoyi:
Will take a look tomorrow

build.sc
```diff
@@ -94,6 +95,13 @@ trait OsLibModule
 
 trait OsModule extends OsLibModule { outer =>
   def ivyDeps = Agg(Deps.geny)
+  override def compileIvyDeps = T{
+    scalaVersion.zip(super.compileIvyDeps).map{
+      case (scalaVer,superDeps) =>
+        val scalaReflectOpt = if (!ZincWorkerUtil.isDottyOrScala3(scalaVer)) Some(Deps.scalaReflect(scalaVer)) else None
+        superDeps ++ Agg.from(scalaReflectOpt)
+    }
+  }
```
lihaoyi, line 104: This is probably better written as

```scala
  override def compileIvyDeps = T{
    val scalaReflectOpt = if (!ZincWorkerUtil.isDottyOrScala3(scalaVersion())) Some(Deps.scalaReflect(scalaVersion())) else None
    super.compileIvyDeps() ++ Agg.from(scalaReflectOpt)
  }
```

os/test/src/PathTests.scala
```diff
@@ -2,16 +2,90 @@ package test.os
...
+
+      testSegmentsFromString("""ahs/""", "ahs" :: "" :: Nil)
+      testSegmentsFromString("""ahs//""", "ahs" :: "" :: "" :: Nil)
+
+      testSegmentsFromString("""ahs/aa/""", "ahs" :: "aa" :: "" :: Nil)
+      testSegmentsFromString("""ahs/aa//""", "ahs" :: "aa" :: "" :: "" :: Nil)
+
+      testSegmentsFromString("""/a""", "" :: "a" :: Nil)
+      testSegmentsFromString("""//a""", "" :: "" :: "a" :: Nil)
+      testSegmentsFromString("""//a/""", "" :: "" :: "a" :: "" :: Nil)
```
lihaoyi, line 42: Let's use the `List(...)` syntax here rather than `::` and `Nil`, for better consistency with the rest of the codebase

os/test/src/PathTests.scala
```diff
@@ -2,16 +2,90 @@ package test.os
...
-
 import os._
-import os.Path.{driveRoot}
+import os.Path.driveRoot
+import os.PathChunk.segmentsFromString
 import utest.{assert => _, _}
+
 import java.net.URI
 object PathTests extends TestSuite {
+  private def nonValidPathSegment(chars: String) = s"[$chars] is not a valid path segment."
```
lihaoyi, line 12: I think we can do a more use-focused error message here, something like 

> literal path sequence <entire-literal> used in OS-Lib cannot have redundant slashes, please use <literal-without-leading-slashes> instead

os/test/src/PathTests.scala
```diff
@@ -2,16 +2,90 @@ package test.os
...
+        compileError { """ root / "src/"  """ }.check("", nonValidPathSegment(""))
+        compileError { """ root / "src/." """ }.check("", nonValidPathSegment("."))
+        compileError { """ root / "src/.." """ }.check("", nonValidPathSegment(".."))
+
+        compileError { """ root / "" """ }.check("", nonValidPathSegment(""))
+        compileError { """ root / "." """ }.check("", nonValidPathSegment("."))
+        compileError { """ root / ".." """ }.check("", nonValidPathSegment(".."))
+
+        compileError(""" root / "hello" / ".." / "world" """).check("", nonValidPathSegment(".."))
+        compileError(""" root / "hello" / "../world" """).check("", nonValidPathSegment(".."))
```
lihaoyi, line 81: Should we allow these scenarios? Allowing literals to generate `ups` seems reasonable, given that they are literal strings, as long as they care canonicalized

lihaoyi:
@pawelsadlo looks great, left some comments. If you update the code and update the PR description to have the most up to date state of the PR I think we can merge it

os/test/src/PathTests.scala
pawelsadlo: I think it's ok, it gives lots of flexibility to users.

We should decide what to do with following edge cases:
`root / ".." / foo`
`root / "../.." / foo`
etc.
 
I think for the first one we should fail with compile error saying to use `up` instead of `".."`.
What do you think?

lihaoyi: Let's allow `root / ".." / foo`. It definitely duplicates `/ os.up /`, but both more concise and arguably more intuitive. And the compile-time-only nature of the conversions should stop the normal problems of untrusted user inputs causing filesystem traversal problems

os/test/src/PathTests.scala
pawelsadlo: @lihaoyi
I have provided a `NonCanonicalLiteral` exception for this case
actually there is also an edge case when sanitizing literal results in empty path, for example `//`, `/.`, `.` literals , in such cases i throw:
```scala
InvalidSegment(
      literal,
      s"Literal path sequence [$literal] doesn't affect path being formed, please remove it"
    )
``` 
If you have any better suggestions for error messages I can change them.

os/test/src/PathTests.scala
pawelsadlo: Actually I am thinking of making this function private[os] , in order not to expose it to public API. 

This means dropping tests, but keeping it public will force us to keep binary compatibility, and can result in users using it to form a paths (which we probably dont want)

lihaoyi: can we move the tests into the same package? I know the tests are mosty in a different package for now, but i dont think that's required

pawelsadlo:
@lihaoyi could you please rerun checks?

lihaoyi:
@pawelsadlo looks great, thanks! I merged it, I assume your bank details haven't changed so I'll pay out the bounty to the same account as last time

PR #298, opened 2024-09-07
Make `os.pwd` modifiable via the `os.pwd0` dynamic variable

lihaoyi:
Basically the same thing we did in https://github.com/com-lihaoyi/os-lib/pull/295 and https://github.com/com-lihaoyi/os-lib/pull/283, except rather than being `os.SubProcess`-specific this applies to all usage of `os.pwd`. And with similar limitations: this works for code paths that go through OS-Lib, but not for code that uses `java.io` or `java.nio` directly.

AFAIK this is the last of the "global" things we need to redirect in Mill tasks, after env variables and default subprocess streams. I'd like it to go into Mill 0.12.0 together with the other sandboxing-related changes

Added unit tests and updated the documentation

Pull Request: https://github.com/com-lihaoyi/os-lib/pull/298

### DIFF
--- a/os/src-jvm/package.scala
+++ b/os/src-jvm/package.scala
@@ -39,9 +39,9 @@ package object os {
   /**
    * The current working directory for this process.
    */
-  val pwd: Path = pwd0.value
+  def pwd: Path = pwd0.value
   val pwd0: DynamicVariable[Path] =
-    DynamicVariable(os.Path(java.nio.file.Paths.get(".").toAbsolutePath))
+    new DynamicVariable(os.Path(java.nio.file.Paths.get(".").toAbsolutePath))
 
   val up: RelPath = RelPath.up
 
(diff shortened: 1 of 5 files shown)


os/src-jvm/package.scala
```diff
@@ -38,7 +39,9 @@ package object os {
   /**
    * The current working directory for this process.
    */
-  val pwd: Path = os.Path(java.nio.file.Paths.get(".").toAbsolutePath)
+  def pwd: Path = pwd0.value
+  val pwd0: DynamicVariable[Path] =
```
lefou, line 43: Can we find a better name? `pwdContext` or `pwdDynamic`?

Would be nice, if reading code gives a clue what's happening.

Example:

```scala
val y = os.pwd0.withValue(os.pwd / "hello") {
  os.pwd
}
```

vs.

```scala
val y = os.pwdContext.withValue(os.pwd / "hello") {
  os.pwd
}
```

lolgab (approved):

os/src-jvm/package.scala
lihaoyi: I went with `dynamicPwd`

PR #316, opened 2024-09-29
Feature implementation for zip & unzip functionality

chaitanyawaikar:
This PR addresses the [issue](https://github.com/com-lihaoyi/os-lib/issues/161) and adds the following functionalities

1. Zipping of files and folders
2. Unzip files
3. Unit test cases covering all the zip and unzip functionality
4. Documentation support for `zip` and `unzip` functions

**Summary of Changes:**

**Zip Functionality:**

- Added options to include (-i), exclude (-x), and delete (-d) files from the zip archive using file name patterns (provided as List[String]).
- Converted exclusion and inclusion patterns from String to Regex for pattern matching.
- Enhanced the zipping logic to support:
      - Excluding files/folders that match specific patterns.
      - Including only specified files.
      - Deleting files that match patterns from an existing zip.
- Added support for appending files to an existing zip archive.

**Unzip Functionality**

- Added support for excluding specific files or folders during extraction using patterns (-x option).
- Implemented a listing feature (-l option) that lists the contents of the zip archive without extracting files.

PR #317, opened 2024-10-07
`zip` and `unzip` API

lihaoyi:
Pulls in changes from https://github.com/com-lihaoyi/os-lib/pull/316 and https://github.com/com-lihaoyi/os-lib/pull/310 and cleans it up

Mostly documented in the readme.adoc. 

Major APIs added:

- `os.zip`: create or append to an existing zip file on disk
- `os.zip.stream`: create a new zip file but write it to an `java.io.OutputStream` rather than a file on disk
- `os.unzip`: unzip a zip file on disk into a folder on disk
- `os.unzip.stream`: unzip a zip file from an `java.io.InputStream` into a folder on disk
- `os.unzip.list`: list the contents of a zip file
- `os.unzip.streamRaw`: low-level API used by `os.unzip.stream` and `os.unzip.list`, exposed in case users need it
- `os.zip.open`: Opens a zip file as `java.nio.file.FileSystem` and gives you an `os.Path` you can use to work with it

Hopefully these are APIs we can start using in Mill rather than `"zip"` subprocesses or ad-hoc helpers like `IO.unpackZip`

Limitations:

* Use of `java.nio.file.FileSystem` is only supported on JVM and not on Scala-Native, and so using `os.zip` to append to existing jar files or `os.zip.open` does not work on Scala-Native. 
* Also `os.zip` doesn't support creating/unpacking symlinks or preserving filesystem permissions in Zip files, because the underlying `java.util.zip.Zip*Stream` doesn't support them. Apache Commons Compress can work with them (https://commons.apache.org/proper/commons-compress/zip.html), but if we're sticking with std lib we don't have that
* Bumps the version requirement to Java 11 and above, matching the direction of the rest of com-lihaoyi. Probably not strictly necessary, but we have to do it eventually and now is as good a time as ever with requests already bumped and Mill bumping soon in 0.12.0


lihaoyi:
@lefou @lolgab could use a review on this if you have time before I merge this, since this is a new API in a pretty wide design space.

.github/workflows/run-tests.yml
```diff
@@ -13,13 +13,8 @@ jobs:
...
       matrix:
-        os: [ubuntu-latest, windows-latest]
-        java-version: [8, 17]
-        include:
-          - os: macos-latest
-            java-version: 17
-          - os: macos-latest
-            java-version: 11
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        java-version: [11, 17]
```
lefou, line 17: Why is that? Is this just convenient or necessary. I'd like to avoid dropping support for a Java version in a minor release. If we need to drop it, we should bump to `0.11.0`. Keeping bin-compat with `0.10.x` is a bonus feature.

lefou (commented):
I think it looks good. OS-Lib is still in version `0` phase, so we can change API of new features rather easy. Let's release it so we get feedback from users.

Why was Java 8 dropped in CI?

.github/workflows/run-tests.yml
lihaoyi: Just convenience really, but i realized we already have `os.Internal.transfer` so I can just use that just as conveniently

lihaoyi: Seems like there's some issue with zip filesystem on windows java 8, so need to bump that to java 11 as well

lihaoyi: I think given that we now have 2/3 OSs only testing on Java 11, let's just bump the official supported version to 11. People can continue to try using it on 8, it'll just be at their own risk

Issue #318, opened 2024-10-07
os.pwd0 unexpected value when compiled with GraalVM and class init at build time.


mio-19:
I built my scala program with GraalVM and class init at build time. From the error message I find that os.pwd0 is defined as a `val`, which was probably calculated by GraalVM compiler at build time, which leads to unexpected behaviour.

lihaoyi:
This seems like a problem with your graal confifg, not with os-lib. We can't change everything to lazy vals to satisfy graal

mio-19:
How about make it a def like https://github.com/com-lihaoyi/os-lib/pull/239?

lihaoyi:
I'd like a deeper investigation before we make another change. Graal has to support cases like this. Static `val`s that require environmental data during initialization are not rare.

If not, I'd like a link to the upstream ticker and discussion in the graal issue tracker before we make any further changes in OS-Lib

mio-19:
One option is to provide native image configuration in META-INF, like what some popular java libraries did

lihaoyi:
I'm not an expert in graal, so you will need to present the problem space and potential solutions

lihaoyi:
Sounds like you need to configure the `--initialize-at-runtime` flag appropriately https://www.graalvm.org/latest/reference-manual/native-image/guides/specify-class-initialization/

Ostrzyciel:
Hi @lihaoyi, we've run into the same issue.

Library vendors can embed GraalVM Native options in their JARs, so that users don't have to configure this manually: https://www.graalvm.org/latest/reference-manual/native-image/overview/BuildConfiguration/

I leave this for your consideration :) this caught us off-guard.

PR #321, opened 2024-10-10
convert to lazy val to fix a problem when compiled with GraalVM Nativ…

mio-19:
…e image and class init at build time

https://github.com/com-lihaoyi/os-lib/issues/318


lihaoyi:
Probably going to call this a wontfix

j-mie6:
How come? is it a binary break?

SethTisue:
@j-mie6 see discussion on #318

j-mie6:
Fair enough!

PR #324, opened 2024-10-20
Add flag to register shutdown hooks for `os.call` and `os.spawn` APIs, overhaul `destroy` APIs

lihaoyi:
* Backports the functionality from Mill to allow us to register shutdown hooks such that when the parent process terminates the subprocesses are shut down as well. This allows the same logic to be used consistently across all `.call` and `.spawn` invocations

* Consolidate `SubProcess#destroy` and `SubProcess#destroyForcibly` into a single `SubProcess#destroy` method which takes some default parameters, allowing the user to choose `async = true` or configure the `shutdownGracePeriod: Long`.

* The default behavior of `SubProcess#destroy` has changed to block on the subprocess actually exiting, which I think is more intuitive. The old behavior is available under `destroy(async = true)`

Best reviewed with `Hide whitespace`. Added some simple unit tests to assert the new functionality, existing unit tests should assert on existing workflows

### DIFF
--- a/Readme.adoc
+++ b/Readme.adoc
@@ -1727,7 +1727,7 @@ os.call(cmd: os.Shellable,
         check: Boolean = true,
         propagateEnv: Boolean = true,
         shutdownGracePeriod: Long = 100,
-        shutdownHook: Boolean = true): os.CommandResult
+        destroyOnExit: Boolean = true): os.CommandResult
 ----
 
 _Also callable via `os.proc(cmd).call(...)`_
@@ -1857,7 +1857,7 @@ os.spawn(cmd: os.Shellable,
          mergeErrIntoOut: Boolean = false,
          propagateEnv: Boolean = true,
          shutdownGracePeriod: Long = 100,
-         shutdownHook: Boolean = true): os.SubProcess
+         destroyOnExit: Boolean = true): os.SubProcess
 ----
 
 _Also callable via `os.proc(cmd).spawn(...)`_
--- a/Readme.adoc
+++ b/Readme.adoc
@@ -1725,7 +1725,9 @@ os.call(cmd: os.Shellable,
         mergeErrIntoOut: Boolean = false,
         timeout: Long = Long.MaxValue,
         check: Boolean = true,
-        propagateEnv: Boolean = true): os.CommandResult
+        propagateEnv: Boolean = true,
+        shutdownGracePeriod: Long = 100,
+        shutdownHook: Boolean = false): os.CommandResult
 ----
 
 _Also callable via `os.proc(cmd).call(...)`_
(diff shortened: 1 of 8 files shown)


lefou (commented):
I think we should rename `shutdownHook` to something like `exitWithJvm` or `stopWithJvm`. We don't really support any custom shtudown hook. We just use the shutdown hook mecahnism internally to stop the process when the JVM is stopped.

lihaoyi:
Ended up going with `destroyOnExit`, to match the existing `os.temp(deleteOnExit = true)` and `.destroy()` methods

Readme.adoc
```diff
@@ -1725,7 +1725,9 @@ os.call(cmd: os.Shellable,
         mergeErrIntoOut: Boolean = false,
         timeout: Long = Long.MaxValue,
         check: Boolean = true,
-        propagateEnv: Boolean = true): os.CommandResult
+        propagateEnv: Boolean = true,
+        shutdownGracePeriod: Long = 100,
```
lefou, line 1729: What's the time unit used for `shutdownGracePeriod`? Should we name it `shutdownGracePeriodMsec`?

os/src/ProcessOps.scala
```diff
@@ -68,7 +99,34 @@ object spawn {
       stdout = stdout,
       stderr = stderr,
       mergeErrIntoOut = mergeErrIntoOut,
-      propagateEnv = propagateEnv
+      propagateEnv = propagateEnv,
+      shutdownGracePeriod = shutdownGracePeriod,
+      destroyOnExit = destroyOnExit
+    )
+  }
+  def apply(
```
lefou, line 107: Is this only provided for bin-compat? Shouldn't we deprecate it therefore?

os/src/ProcessOps.scala
```diff
@@ -40,7 +38,38 @@ object call {
       timeout = timeout,
       check = check,
       propagateEnv = propagateEnv,
-      timeoutGracePeriod = timeoutGracePeriod
+      shutdownGracePeriod = shutdownGracePeriod,
+      destroyOnExit = destroyOnExit
+    )
+  }
+  def apply(
```
lefou, line 45: Looks like this one is only for bin-compat, so we should deprecate it for removal.

Readme.adoc
lihaoyi: I'll probably leave it as is, the scaladoc says it's in milliseconds so that should be enough

os/src/ProcessOps.scala
lihaoyi: I've had issues with `@deprecated` things being accidentally resolved here before, but will add a comment like the other forwarders

PR #325, opened 2024-10-23
Instrumented path based operations using hooks defined in `Checker`

ajaychandran:
Instrumented path based operations using hooks defined in `Checker`.
```scala
trait Checker {
  def onRead(path: ReadablePath): Unit
  def onWrite(path: Path): Unit
}
```

### Exceptions
The following operations were not instrumented:
- `followLink`, `readLink`
- `list`, `walk`
- `exists`, `isLink`, `isFile`, `isDir`
- read operations for permissions/stats
- `watch`

### Future work
- A more comprehensive design would add hooks for each core operation. This would eliminate the special check handling in operations like `move` and `symlink`.
- As such, the methods of `ReadablePath` represent escape hatches. These cannot be "plugged" without breaking binary compatibility.

This resolves part 1 of [mill #3746](https://github.com/com-lihaoyi/mill/issues/3746).


os/src/FileOps.scala
```diff
@@ -191,18 +201,17 @@ object copy {
...
-        target.wrapped
       } else {
         Files.copy(p.wrapped, target.wrapped, opts1 ++ opts2 ++ opts3: _*)
       }
     }
 
     copyOne(from)
-    if (stat(from, followLinks = followLinks).isDir) walk(from).map(copyOne)
+    if (stat(from, followLinks = followLinks).isDir) for (p <- walk(from)) copyOne(p)
   }
```
ajaychandran, line 215: A minor optimization that eliminates the creation of an unused collection.
This isn't required for the original ticket.

os/test/src/FilesystemPermissionsTests.scala
```diff
@@ -50,5 +74,37 @@ object FilesystemPermissionsTests extends TestSuite {
         }
       }
     }
```
ajaychandran, line 76: The original "group" test was testing `owner` operations.

ajaychandran, line 76: The original "group" test was testing `owner` operations.

os/test/src/FilesystemPermissionsTests.scala
```diff
@@ -34,12 +56,14 @@ object FilesystemPermissionsTests extends TestSuite {
...
         if (Unix()) {
           // Only works as root :(
           if (false) {
+            intercept[WriteDenied] {
+              os.owner.set(rd / "File.txt", "nobody")
+            }
+
             val originalOwner = os.owner(wd / "File.txt")
 
             os.owner.set(wd / "File.txt", "nobody")
```
ajaychandran, line 69: The original "group" test was testing `owner` operations.

os/src/FileOps.scala
```diff
@@ -191,18 +201,17 @@ object copy {
...
         // nothing to do
-        target.wrapped
       } else {
         Files.copy(p.wrapped, target.wrapped, opts1 ++ opts2 ++ opts3: _*)
       }
     }
 
     copyOne(from)
-    if (stat(from, followLinks = followLinks).isDir) walk(from).map(copyOne)
+    if (stat(from, followLinks = followLinks).isDir) for (p <- walk(from)) copyOne(p)
```
ajaychandran, line 214: A minor optimization that eliminates the creation of an unused collection.
This isn't required for the original ticket.

os/src/ZipOps.scala
```diff
@@ -45,6 +45,9 @@ object zip {
       deletePatterns: Seq[Regex] = List(),
       compressionLevel: Int = java.util.zip.Deflater.DEFAULT_COMPRESSION
   ): os.Path = {
+    checker.value.onWrite(dest)
+    // check read preemptively in case "dest" is created
+    for (source <- sources) checker.value.onRead(source.src)
```
ajaychandran, line 50: Alternatively, the implementation should be refactored to defer file creation until all data is collected.

ajaychandran:
@lihaoyi Please review.

lihaoyi:
Thanks @ajaychandran, I think this looks good. Could you add Scaladoc to the `checker` API as well? Then I can merge this and cut an unstable `-M1` so you can start using it in Mill

lihaoyi:
> The following operations were not instrumented:

Let's instrument these as well. For the purposes of Mill's (and probably others), we also want to block code from making decisions based on filesystem metadata outside their designated sandboxes, not just file contents

ajaychandran:
> Could you add Scaladoc to the checker API as well?

I had removed Scaladoc from the PR description. Do we need more than this?
```scala
/**
 * Defines hooks for path based operations.
 *
 * This, in conjunction with [[checker]], can be used to implement custom checks like
 *  - restricting operations to some path(s)
 *  - logging operations
 */
trait Checker {

  /** A hook for a read operation on `path`. */
  def onRead(path: ReadablePath): Unit

  /** A hook for a write operation on `path`. */
  def onWrite(path: Path): Unit
}```

**Comment by ajaychandran** [comment]
@lihaoyi 
The current implementation assumes an equivalence between write and create/remove operations. I am not sure if this would fit all use cases.
- The modified implementation for `move` checks for write access on the source parent directory. I gather this is not per your expectation.
- I assumed a `symlink`ed path could be used as argument to read. If so, a user could gain access to any file.
- Should `makeDir` test write access on an existing parent folder when `createFolders` is true?

Adding a hook for each core operation lets `os-lib` remain unopinionated.
Maybe rename `Checker` to `Hook`?

File: os/src/FileOps.scala [hunk 6]
```diff
@@ -38,6 +42,7 @@ object makeDir extends Function1[Path, Unit] {
   object all extends Function1[Path, Unit] {
     def apply(path: Path): Unit = apply(path, null, true)
     def apply(path: Path, perms: PermSet = null, acceptLinkedDirectory: Boolean = true): Unit = {
+      checker.value.onWrite(path)
```
dabd, line 45: The original implementation creates any necessary enclosing directories via `java.nio.file.Files#createDirectories`.  

Here the checker is only checking for write access in the full path.

Don't we need to apply it to the enclosing directories as well?

I see this would entail providing a shim for `java.nio.file.Files#createDirectories`.

lihaoyi:
Good point on the symlinks and enclosing folders. In the end this is not going to be an airtight sandbox, so it's a judgement call exactly where to draw the line.

1. For `move`, let's call `onWrite` it on the full source and destination paths  

2. I think for creating enclosing folders, let's just punt on that and say it's up the the implementer of `def onWrite`

3. For symlinks, I think what you say is reasonable, so lets call `onWrite` on the destination path, since we can compute it in-memory by resolving any relative path without needing to perform any filesystem operations

I'm a bit reluctant to try and match the entire OS-Lib API in the `Checker` interface. Let's go with the more minimal ad-hoc interface for now. 

For the purposes of this pull request, could you add an `@experimental` annotation, apply it to the stuff you added, and add it to the `def mimaExcludeAnnotations` in the build file? Just to make it clear that this API isn't stable, and we're only publishing it because we need to experiment with it in Mill

lihaoyi:
@ajaychandran not everything needs to be experimental, only the public APIs you added: `os.checker` and `os.Checker` I think

ajaychandran:
@lihaoyi This is ready for another review.

os/src-jvm/package.scala
```diff
@@ -59,6 +59,8 @@ package object os {
 
   val sub: SubPath = SubPath.sub
 
+  val checker: DynamicVariable[Checker] = new DynamicVariable[Checker](Checker.Nop)
```
dabd, line 62: How about adding these for ergonomics?

```scala
def withChecker[T](checker: Checker)(op: => T): T = {
    checker.withValue(checker)(op)
  }
  
def checkReadAccess(path: os.Path): Unit = checker.value.onRead(path)
def checkWriteAccess(path: os.Path): Unit = checker.value.onWrite(path)
```

lihaoyi: I think for now we can punt on these, since it's a pretty advanced feature so being easy to use isn't as important

ajaychandran: Whats the benefit here?
The number of characters typed at the call site is roughly the same.

os/src/FileOps.scala
dabd: Nice optimization! Would `walk(from).foreach(copyOne)` be more idiomatic?

ajaychandran: I was trying to avoid [this](https://github.com/com-lihaoyi/mill/pull/3756#discussion_r1807286381).

lihaoyi:
@ajaychandran code changes look good. Last request here is to move all the checker tests into one separate test suite, rather than having them mixed in with all the non-checker tests. That will make it easier to skim them and audit them as necessary.

After that i will merge this and cut a milestone release for experimentation in Mill

ajaychandran:
@lihaoyi Checker tests have been moved.

Some compound operations like `zip` incur the cost of double-checking.
One solution would be to define a `private[os] def unchecked` variant, for each operation, for internal use. This can be deferred until the feature is stable.

dabd:
> @lihaoyi Checker tests have been moved.
> 
> Some compound operations like `zip` incur the cost of double-checking. One solution would be to define a `private[os] def unchecked` variant, for each operation, for internal use. This can be deferred until the feature is stable.

Another option is to delegate some sort of caching solution to the implementor of Checker.

lihaoyi:
@ajaychandran pushed a tag for 0.11.4-M1 https://github.com/com-lihaoyi/os-lib/actions/runs/11537940922 hopefully it goes out soon and you can try it out in Mill. If you're going to work on the downstream tasks, we can hold off on payment for now to save on transfer fees. If not I can transfer the first portion of the bounty using the details we used before

ajaychandran:
@lihaoyi Thanks. Lets hold off the transfer.

Issue #327, opened 2024-10-29
`os.pwd / "../folder/../Out.txt"` seems to be allowed despite being a non-canonical sub-path


lihaoyi:
Seems like we should error and ask people to us `"../Out.txt"`? @pawelsadlo what do you think?

pawelsadlo:
@lihaoyi 
I am not sure how would it behave in case folder is symlink.
We should check what is behavior in case of symlink without literal.

lihaoyi:
Mill already normalizes `os.pwd / ".." / "folder" / ".." / "Out.txt"` `to os.pwd / ".." / "Out.txt"` on-creation, even before we perform any filesystem operations, so whether symlinks behave correctly or not, those are the semantics we already use

PR #329, opened 2024-11-02
Add compile-time validation of literal paths containing ".."

pawelsadlo:
Addresses #327 

This PR adds additional validation of literal path segments containing `..`.

It throws compiletime error, suggesting canonical form of a path, when literal path is not canonical.

Eg. "../foo/../bar"   suggests  "../bar"

"foo/.." suggests removing literal


lihaoyi:
Thanks @pawelsadlo !

PR #330, opened 2024-11-14
Fix empty folder missing problem when zip files.

counter2015:
Fix: #328

PR #350, opened 2025-01-31
Revert "Fix empty folder missing problem when zip files. (#330)"

### LINKED ISSUE #349: zip is zipping up the source folder instead of its content
Happens only in `0.11.4-M3` and `0.11.4-M4`.
`0.11.3` or `0.11.4-M2` work, so it's in between M2 and M3.

# Reproduce

```sh
mkdir os-lib-issue-x && cd $_

curl -o scala https://raw.githubusercontent.com/VirtusLab/scala-cli/main/scala-cli.sh && chmod +x scala

mkdir sources
echo "toto1" > sources/file1.txt
echo "toto2" > sources/file2.txt

cat > run.scala <<EOF
//> using scala 3.6.3
//> using dep com.lihaoyi::os-lib:0.11.4-M4

@main def run() =
  val sources = os.pwd / "sources"
  val myZip = os.pwd / "my.zip"
  os.remove(myZip)
  os.zip(myZip, Seq(sources))

  val content = os.unzip.list(myZip).toSet
  val expected = os.list(sources).map(_.relativeTo(sources)).toSet
  assert(content == expected, s"\$content != \$expected")
EOF

./scala .
```

Thanks

joan38:
This reverts commit 90626dfd9f4371c8eb88ff794faefe7a841af90d.
Fixes #349

I've spent some time trying to do a fix forward but I don't have much time on my hand.
Can we revert this in the meantime and @counter2015 can provide another PR for the empty folder issue?

### DIFF
--- a/os/src/ZipOps.scala
+++ b/os/src/ZipOps.scala
@@ -62,7 +62,7 @@ object zip {
           excludePatterns,
           includePatterns,
           (path, sub) => {
-            os.copy.over(path, opened / sub, createFolders = true)
+            os.copy(path, opened / sub, createFolders = true)
             if (!preserveMtimes) {
               os.mtime.set(opened / sub, 0)
               // This is the only way we can properly zero out filesystem metadata within the
@@ -98,12 +98,10 @@ object zip {
     sources.foreach { source =>
       if (os.isDir(source.src)) {
         for (path <- os.walk(source.src)) {
-          if (shouldInclude(path.toString, excludePatterns, includePatterns)) {
-            val target = source.dest.getOrElse(os.sub) / path.subRelativeTo(source.src / os.up)
-            makeZipEntry0(path, target)
+          if (os.isFile(path) && shouldInclude(path.toString, excludePatterns, includePatterns)) {
+            makeZipEntry0(path, source.dest.getOrElse(os.sub) / path.subRelativeTo(source.src))
           }
         }
-        makeZipEntry0(source.src, source.dest.getOrElse(os.sub / source.src.last))
       } else if (shouldInclude(source.src.last, excludePatterns, includePatterns)) {
         makeZipEntry0(source.src, source.dest.getOrElse(os.sub / source.src.last))
       }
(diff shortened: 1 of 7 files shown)


joan38:
@lihaoyi can you rerun the failing build?

joan38:
Ah ok thanks

lihaoyi:
Tagged this as 0.11.4-M5, will propagate it to Mill once it's published

counter2015:
@joan38 Thanks for report.  I'm not sure whether I understand correctly. It seems the commit break the unwritten rules. But in my persion thinking, the newer one is more intuitive. The former version seems handle folder and file differently. Do you except that when user pass a folder to zip parameter, we should access inside it ?

Then I have a quesion here, if user want to keep the outer folder, how can we do that?
In first glance, I think we can pass the parent folder but exclude other content (it's not convenient).

The common `zip` command in linux support pass muti source include files and folders and keep the outside

For example
```
$ tree
.
├── run.scala
└── sources
    ├── file1.txt
    └── file2.txt


$ zip archieve.zip sources run.scala
  adding: sources/ (stored 0%)
  adding: run.scala (deflated 41%)

$ tree
.
├── archieve.zip
├── my.zip
├── run.scala
└── sources
    ├── file1.txt
    └── file2.txt

```

You could see that the `sources` folder is keeping.

I acknowledge that my previous PR may have introduced breaking changes due to oversights in the implementation. However, from my perspective, retaining directory structures during compression aligns more closely with user intuition.  What's your opinion ? @lihaoyi

joan38:
@counter2015 The readme says:
> This will create a new zip archive at dest containing file1.txt and everything inside sources.

But your point is fair and I thought about that too.

I'll give you an example taken from what broke downstream for a Mill plugin:
```scala
os.zip(dest / "docusaurus.zip", Seq(wd / "build"))
```
How would you implement this with your implementation?
My fix was:
```scala
os.zip(dest / "docusaurus.zip", os.list(wd / "build").map(os.zip.ZipSource.fromPath))
```
(BTW I think we need some conversions for `Seq[Path]` to `Seq[ZipSource]` to avoid `.map(os.zip.ZipSource.fromPath)`)

Both are good to me but maybe we need a `0.12.0` version bump for such change? @lihaoyi WDYT?

Also I want to thank you for your contribution @counter2015, it's very much appreciated.

Thanks

joan38:
I just opened https://github.com/com-lihaoyi/os-lib/pull/352
Maybe with that we could make the change but not sure about the breaking aspect of it.

PR #368, opened 2025-03-05
Fix os.write.outputStream with perms for Scala 3

kiendang:
On Scala 3

```scala
os.write.outputStream(os.pwd / "text", perms = os.PermSet(420))

java.lang.ClassCastException: class [Ljava.nio.file.attribute.FileAttribute; cannot be cast to class scala.collection.immutable.Seq ([Ljava.nio.file.attribute.FileAttribute; is in module java.base of loader 'bootstrap'; scala.collection.immutable.Seq is in unnamed module of loader java.net.URLClassLoader @4520ebad)
  os.write$.outputStream(ReadWriteOps.scala:35)
  ammonite.$sess.cmd0$.<clinit>(cmd0.sc:1)
```

### DIFF
--- a/os/src/ReadWriteOps.scala
+++ b/os/src/ReadWriteOps.scala
@@ -30,8 +30,8 @@ object write {
     checker.value.onWrite(target)
     if (createFolders) makeDir.all(target / RelPath.up, perms)
     if (perms != null && !exists(target)) {
-      val permArray =
-        if (perms == null) Array[FileAttribute[PosixFilePermission]]()
+      val permArray: Array[FileAttribute[_]] =
+        if (perms == null) Array.empty
         else Array(PosixFilePermissions.asFileAttribute(perms.toSet()))
       java.nio.file.Files.createFile(target.toNIO, permArray: _*)
     }
--- a/os/test/src/ReadingWritingTests.scala
+++ b/os/test/src/ReadingWritingTests.scala
@@ -124,6 +124,19 @@ object ReadingWritingTests extends TestSuite {
           os.read(wd / "New File.txt") ==> "Hello"
         }
       }
+      test("outputStreamWithPerms") {
+        test - prep { wd =>
+          val out = os.write.outputStream(wd / "New File.txt", perms = os.PermSet(420))
+          out.write('H')
+          out.write('e')
+          out.write('l')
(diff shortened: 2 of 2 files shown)


lihaoyi:
Looks reasonable, but do you know why it only fails on Scala 3? I would expect it behave the same on all versions of Scala

os/src/ReadWriteOps.scala
```diff
@@ -30,8 +30,8 @@ object write {
     checker.value.onWrite(target)
     if (createFolders) makeDir.all(target / RelPath.up, perms)
     if (perms != null && !exists(target)) {
-      val permArray =
-        if (perms == null) Array[FileAttribute[PosixFilePermission]]()
```
kiendang, line 34: @lihaoyi It's because the type here is wrong. It should be `Array[FileAttribute[java.util.Set[PosixFilePermission]]]` instead. I don't know how this compiles or why it runs on Scala 2 though.

kiendang: I don't know the details, but looks like Scala 3 handles varargs differently https://docs.scala-lang.org/scala3/reference/changed-features/vararg-splices.html

kiendang:
Also looks like on Scala 2.13 using `: _*` with `Array` is a bit inefficient

```scala
val arr = Array(0, 1, 2, 3)
val lst = List(arr: _*)

Passing an explicit array value to a Scala varargs method is deprecated (since 2.13.0) and will result in a defensive copy; Use the more efficient non-copying ArraySeq.unsafeWrapArray or an explicit toIndexedSeq call
val lst = List(arr: _*)
```

Looks Scala 3 doesn't have the problem, but have to use the new syntax unavailable in Scala 2

```scala
val arr = Array(0, 1, 2, 3)
val lst = List(arr*) 
```

For Scala 2, these would not lead to copying

```scala
val arr = Seq(0, 1, 2, 3)
val lst = List(arr: _*)
```

```scala
import scala.collection.immutable.ArraySeq.unsafeWrapArray

val arr = Seq(0, 1, 2, 3)
val lst = List(unsafeWrapArray(arr): _*)
```

lihaoyi:
Thanks @kiendang !

PR #369, opened 2025-03-05
Support zipping empty directories

kiendang:
Currently empty directories are excluded when zipping.

### DIFF
--- a/os/src/ZipOps.scala
+++ b/os/src/ZipOps.scala
@@ -99,7 +99,7 @@ object zip {
       if (os.isDir(source.src)) {
         val contents = os.walk(source.src)
         if (contents.isEmpty)
-          makeZipEntry0(source.src, source.dest.getOrElse(os.sub / source.src.last))
+          source.dest.foreach(makeZipEntry0(source.src, _))
         for (path <- contents) {
           if (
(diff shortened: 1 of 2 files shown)


lihaoyi:
So my understanding of the expected behavior from the current docs (https://github.com/com-lihaoyi/os-lib#oszip)  is that folders directly referenced in the  `zip` call's `sourcePaths` are unpacked into the root of the zip by default, unless they are given an explicit destination path in the zip via `-> os.sub / "renamed-folder"`

In that case, ignoring empty directories as shown in your test case `os.zip(sources = Seq(emptyDir, outerEmptyDir))` seems like the right thing to do, and this change would make it inconsistent with the behavior for non-empty directories

kiendang:
I see. Updated the PR to handle zipping empty directories 

- specified in `sources`
  - without `dest` -> ignored
  - with `dest` -> included
-  inside one of the `sources` -> included

In cases that empty directories are included in the zip, their names should also follow `includePatterns`/`excludePatterns`?

lihaoyi:
@kiendang good question about the `includePatterns`/`excludePatterns`. What does the unix `zip` command do?

kiendang:
Yup the `zip` command on Mac does support `-x`/`-i` for empty directories. Updated.

Issue #356, opened 2025-02-14
Properly support permissions and symlinks in Zip files (500USD Bounty)


lihaoyi:
---

**From the maintainer Li Haoyi**: I'm putting a 500USD bounty on this issue, payable by bank transfer on a merged PR fixing this.

See https://github.com/orgs/com-lihaoyi/discussions/6 for other bounties and the terms and conditions that bounties operate under 

---

@kiendang hit some issues using `os.zip` in Mill (https://github.com/com-lihaoyi/mill/pull/4503), some of which are permissions related. It seems that since 2019 the JDK's zip file APIs should be able to support permissions (https://bugs.java.com/bugdatabase/view_bug?bug_id=8213031), so we should make sure OS-Lib takes advantage to preserve permissions when zipping and unzipping files and folders

kiendang:
Turned out the problem was all permission related. It worked after executable perm was given to files inside `bin` and `libexec`.

Anw can see 2 problems with `os.unzip`: It does not perserve file permissions and symlinks. After being unzipped from `cosmocc.zip` (https://cosmo.zip/pub/cosmocc/) with either `os.unzip` or `os.unzip.stream`, the files are stripped of permissions (files in `bin` no longer have executable permission) and all the original symlinks are replaced with copied files.

sideeffffect:
https://unix.stackexchange.com/a/509337

> Info-Zip 3.0 SUPPORTS preserving files/dirs UNIX permissions and UID/GID ownership data. `zip` stores it by default but you need to use `unzip` in an special way to restore them:
>
> *    `unzip` must be used with the `-X` flag.
> *    `unzip` must run as `root` to set the files/dirs UID/GID. If you run it as a normal user then the UID will be always the one of the current user and the GID will be restored ONLY IF the current user belongs to that group.
> 
> Note: you can also use `unzip` with the `-K` flag to also restore SUID/SGID/Sticky bits.

kiendang:
Extra fields have been used to store information about POSIX permissions and symlink though. `jdk.zipfs` does support POSIX permissions from JDK 14.

`os-lib` uses `java.util.zip.ZipEntry` which does not expose this information though. It's stored in the private `ZipEntry.externalAttributes` field. There's [talk](https://mail.openjdk.org/pipermail/core-libs-dev/2024-May/123001.html) on making it accessible but I don't think it's gonna be resolved any time soon judging on the lack of activity.

[...]

PR #371, opened 2025-03-06
Preserve zip files permissions

sake92:
Seems like we can open the ZIP after it is created, and then add the permissions to the ZIP entries.
See the docs, it needs to be done through java's zip filesystem https://docs.oracle.com/en/java/javase/17/docs/api/jdk.zipfs/module-summary.html

Same with unzip, we extract the files and then dig into the ZIP via zip filesystem to fetch the perms.

Seems like it works.  ¯\\_(ツ)_/¯


sake92:
I've tested it locally with the `cosmocc.zip` from the linked issue in mill, seems to work fine (executables):
```
.......
-rwxrwxr-x 1 sake sake    2461 Mar  6 02:15 cosmoar*
-rwxrwxrwx 1 sake sake       7 Mar  6 02:15 cosmoc++*
-rwxrwxr-x 1 sake sake   18647 Mar  6 02:15 cosmocc*
-rwxrwxr-x 1 sake sake    8667 Mar  6 02:15 cosmocross*
-rwxrwxr-x 1 sake sake    2418 Mar  6 02:15 cosmoinstall*
-rw-rw-r-- 1 sake sake      46 Mar  6 02:15 cosmoranlib
```

os/src/ZipOps.scala
```diff
@@ -255,6 +289,29 @@ object unzip {
...
+          "enablePosixFileAttributes" -> "true",
+          "defaultPermissions" -> Set.empty[String].asJava
+        ).asJava
+      )
+      try {
+        os.walk(dest).foreach { path =>
+          val relPath = path.subRelativeTo(dest).toString
+          if (os.zip.shouldInclude(relPath, excludePatterns, includePatterns)) {
+            val entry = zipFS.getPath(relPath)
+            val permissions = Files.getPosixFilePermissions(entry)
```
kiendang, line 307: I think you should use the more low-level `Files.getAttribute(entry, "zip:permissions")` here. 
In case the zip entry does not have permission information, `Files.getAttribute(entry, "zip:permissions")` would be empty and we could skip `os.perms.set`.

If using `Files.getPosixFilePermissions`, according to the doc

> The "permissions" attribute is not optional in the "posix" view so a default set of permissions are used for entries that do not have access permissions stored in the Zip file. The default set of permissions are
  OWNER_READ
  OWNER_WRITE
  GROUP_READ

sake92: But the unzipped file would get some default permissions even if we dont set them, right?
With this approach we at least can set it with the "defaultPermissions" property.
Currently I set it to empty

sake92: Yeah, you are right, I'm not super familiar with *nix permissions.  
Fixed now, looks good?

sake92:
Closing based on discussion in https://github.com/com-lihaoyi/os-lib/issues/356#issuecomment-2705316204
Seems like it's best to have a separate artifact for zipping.

os/src/ZipOps.scala
kiendang: So I didn't have a correct understanding of umask either. There're no default permissions in Unix. Applications set permissions to files they create. Umask will then mask some of the permissions. What you did before was totally fine. https://askubuntu.com/questions/44542/what-is-umask-and-how-does-it-work

Issue #319, opened 2024-10-07
[Feature request] Scala.js with NodeJS


mio-19:
I don't know if this is out of the scope of this project, but it would be great to  support such environment.

reid-spencer:
[Ossum Inc.](https://gitnub.com/ossuminc) intends to start this work, initially to satisfy just what we need for our product, but with help from others, perhaps the entire API. Our fork is [here](https://github.com/ossuminc/os-lib)

lihaoyi:
@reid-spencer go for it! Feel free to fork the codebase and do whatever you want, once the dust settles we can see where you end up and whether there's a path to upstreaming it into this main repo

PR #372, opened 2025-03-06
Support zip permissions and symlinks with Apache Commons Compress

kiendang:
Use Apache Commons Compress to support symlinks and permissions.

This supports `os.zip` (for new zips, not modification), `os.zip.stream` and `os.unzip`.

Supporting symlinks and permissions for `os.unzip.stream` is impossible since the information is stored in the central directory at the end of the zip file, not together with each zip entry.

`ZipFile` doesn't support modifying entries in place. Thus `os.zip` modification would still require `jdk.zipfs`, which doesn't support symlinks, and only supports permissions since JDK 14.

Apache Commons Compress is quite heavy a dependency since it also includes support for other compressing formats, and also depends on Apache Commons IO. A more light weight alternative would be to vendor `org.apache.tools.zip` from Apache Ant (which is basically the same thing as Apache Commons Compress). See [my other branch](https://github.com/kiendang/os-lib/tree/ant) that goes with this approach.

What's missing?
- [ ] move to `src-jvm`
- [ ] make Apache Commons Compress provided dep?
- [ ] add tests
- [ ] make sure this doesn't break anything on Windows

### DIFF
--- a/build.mill
+++ b/build.mill
@@ -26,6 +26,7 @@ object Deps {
   val sourcecode = ivy"com.lihaoyi::sourcecode::0.4.2"
   val utest = ivy"com.lihaoyi::utest::0.8.4"
   val expecty = ivy"com.eed3si9n.expecty::expecty::0.16.0"
+  val apacheCommonsCompress = ivy"org.apache.commons:commons-compress:1.27.1"
   def scalaReflect(scalaVersion: String) = ivy"org.scala-lang:scala-reflect:$scalaVersion"
   def scalaLibrary(version: String) = ivy"org.scala-lang:scala-library:${version}"
 }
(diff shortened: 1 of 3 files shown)


sake92:
os-lib has(had) a ScalaNative build https://github.com/com-lihaoyi/os-lib/blob/main/build.mill#L197-L204
and in the future it might have a [ScalaJS](https://github.com/com-lihaoyi/os-lib/issues/319) (nodejs) build too.  
So it is preferrable to add apache-commons-compress only to the JVM build.
This means we'd have to introduce interface(s) for ZIP/UNZIP actions, and implement it for each platform.

Maybe something like:
```scala
trait ZipActions {
  def zip(
      dest: os.Path,
      sources: Seq[ZipSource] = List(),
      excludePatterns: Seq[Regex] = List(),
      includePatterns: Seq[Regex] = List(),
      preserveMtimes: Boolean = false,
      deletePatterns: Seq[Regex] = List(),
      compressionLevel: Int = java.util.zip.Deflater.DEFAULT_COMPRESSION
  ): os.Path

  def unzip..
}
```
and then implement it for JVM as `JvmZipActions extends ZipActions` via commons-compress.  
For native we could drop to shelling out or something, whatever we find working in the future...

---
Then we can have for JVM:
```scala
object PlatformActions {
  val zipActions = new JvmZipActions()
}
```
and native
```scala
object PlatformActions {
  val zipActions = new NativeZipActions()
}
```
From `ZipOps.scala` we could call `PlatformActions.zipActions.zip(...)`

PR #373, opened 2025-03-06
fix os.copy with mergeFolders=true

pawelsadlo:
Bug appears when we call os.copy with mergeFolders = true

if target doesn't contain inner directories corresponding to source os.copy with mergeFolders will fail try to copy source directory as it was a file and fail with FileAlreadyExistsException

createFolders = true doesn't fix things

this is because os.isDir(targetDir) returns false if path doesn't exist

This PR adds explicit target directory creation if sourceDirectory is directory and targetDirectory doesn't exist

### DIFF
--- a/os/src/FileOps.scala
+++ b/os/src/FileOps.scala
@@ -202,7 +202,9 @@ object copy {
 
     def copyOne(p: Path): Unit = {
       val target = to / p.relativeTo(from)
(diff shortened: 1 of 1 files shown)


lefou:
Can you please add a test for the case you described?

pawelsadlo:
@lefou Never mind, I misunderstood replaceExisting parameter behaviour (thought it would simply omit coliding files but its actually throwing exception)

PR #374, opened 2025-03-07
Add support for permissions and symlinks in os.zip/unzip with vendored zip source code from Apache Ant

kiendang:
### Vendored code

Follow the discussion in #356, this adds support for permissions and symlinks in `os.zip`/`unzip` with vendored zip source code from Apache Ant.

The vendored source code is generated by the `os.zip.apacheAntZipSource` task and put in `os/zip`. It's shaded with the package renamed from `org.apache.tools.zip` to `os.shaded_org_apache_tools_zip`.

`scala-steward.conf` was added and configured to run `os.zip.apacheAntZipSource` on `org.apache.ant:ant` updates.

### Features

This brings support for permissions and symlinks to `zip` (for creating new zips, not modifying existing ones), `zip.stream` and `unzip`. As for modifying existing zips, we would still have to rely on `jdk.zipfs` which does not support symlinks.

| | file permissions | symlinks |
| --- | --- | --- |
| `os.zip.open` | if Java Runtime Version >= 14 | |
| `os.zip` (create new) | ✅ | ✅ |
| `os.zip` (modify existing) | if Java Runtime Version >= 14 | |
| `os.zip.stream` | ✅ | ✅ |
| `os.unzip` | ✅ | ✅ |
| `os.unzip.stream` | | |

### TODO
- [ ] **(Advice needed)** make sure we comply with Apache Ant's license to include the code here. Would appreciate opinions on this as I'm not an expert.
- [ ] **(Advice needed)** make `ZipOps` JVM only
- [x] tests
- [x] make sure things don't break on Windows
- [x] add permission support to modifying existing zips with `jdk.zipfs` like what @sake92 did in #371


lihaoyi:
At a first glance, how about we change the package name to `os.shaded_org_apache_tool_zip` and keep the class names unchanged. That would reduce the amount of diff we would have to apply to the source files and keep them closer to their upstream equivalents

build.mill
```diff
@@ -216,3 +218,45 @@ object os extends Module {
     }
   }
 }
+
+trait ApacheAntZipVendor extends Module {
```
lihaoyi, line 222: This can probably be a normal `object` rather than a `trait`.

It also deserves a comment summarizing the reasons:

* Why we chose to include a third party implementation (because the JDK version doesn't support symlinks) 
* Why we chose the Ant implementation over apache-commons (to avoid a dependency on apache-commons-io, and anyway it's where the commons-io implementation came from) 
* Why we are shading it (to ensure it doesn't cause classpath conflicts)
* Why we are building it from source (because we only want a subset of the codebase, not the entire thing)

lihaoyi: Let's also compile the shaded apache zip stuff in a `os.zip` submodule in `os/zip/src/` separate from the main OS-Lib codebase in `os/src/`. It's generally best to avoid mixing generated and hand-written code.

lihaoyi:
Needs some small unit tests exercising the permissions/symlink preservation functionality, and a manual test on the `cosmocc` zip to reproduce the problems of the old implementation and verify that it works with the new one

kiendang:
> At a first glance, how about we change the package name to os.shaded_org_apache_tool_zip and keep the class names unchanged. That would reduce the amount of diff we would have to apply to the source files and keep them closer to their upstream equivalents

I would like that too, but Java package private access modifiers are not as flexible as in Scala. You can't have `package os.shaded_org_apache_tool_zip` and make it `private[os]`, only `private[os.shaded_org_apache_tool_zip]` (no modifier in Java).

How about move `ZipOps.scala` to a trait, then move both the trait and the vendored code to a `os.zip` package, make the vendored code `private[os.zip]` and `ZipOps` trait `private[os]` and mix in the `ZipOps` trait with `os` package object? We probably need to move `ZipOps.scala` to a trait anyway to make it JVM only.

lihaoyi:
I think not privatizing it is fine. We can just document it as unstable and turn of MIMA enforcement for it

lihaoyi:
We could also put a shim Scala file in `os.shaded_org_apache_tool_zip` that has access to the rest of the Java code, and give the Scala file `private[os]` so it can then be accessed by the rest of os-lib

kiendang:
> We could also put a shim Scala file in `os.shaded_org_apache_tool_zip` that has access to the rest of the Java code, and give the Scala file `private[os]` so it can then be accessed by the rest of os-lib

I tried it (https://github.com/com-lihaoyi/os-lib/pull/374/commits/05ee56b8b94a2651fb5dc81092322c94eaaf2c4f) but it doesn't work. Looks like we can't bypass Java access modifier that way.

```
<error class ZipOutputStream cannot be accessed as a member of (os.shaded_org_apache_tools_zip : shaded_org_apache_tools_zip) from module class zip$.>
```

Got to either keep it public or put `os.zip` under the same package with the vendored code. Currently doing the former.

kiendang:
Added the `followLinks` arg to `os.zip`/`zip.stream`. `true` means zipping symlinks as the referenced files while `false` means zipping them as symlinks. It's default to `true`, similar to the `zip` command on Unix (you have to pass `-y` to store symlinks as symlinks). Plus symlinks don't work on Windows, so storing symlinks as the referenced files is a good default.

`os.zip` on an existing zip file has some caveats related to permissions and symlinks. Added those to docstring.

Manual tested `os.unzip` on `cosmocc.zip` and it worked. Test script [here](https://gist.github.com/kiendang/cbe24fc969d9646c9f296f7d8654e1cf).

kiendang:
I moved the shaded code to a separate module in `os/zip/src/`, but that means separate artifact for it? Shouldn't the shaded classes be package in the same `os-lib` jar?

Also I just saw https://github.com/com-lihaoyi/mill/issues/3815. This should be a good use case for that? What's done here should be supported by jarjar-abrams, more specifically the `Rename` and `Keep` rules to rename the package and only keep the zip-related classes from the whole Ant jar.

lihaoyi:
I think separate artifact is fine. OS-Lib already has a few artifacts (e.g. os-lib-watch), and depends on upstream things (e.g. geny)

The shading support could use this, but it's largely unrelated. I imagine it would largely be using jarjar or jarjarlinks or some other third-party tool directly

lihaoyi:
I think this looks reasonable

> This brings support for permissions and symlinks to zip (for creating new zips, not modifying existing ones), zip.stream and unzip. As for modifying existing zips, we would still have to rely on jdk.zipfs which does not support symlinks.

Could you update the readme file to describe the limitations here? e.g. which APIs support symlinks and permissions, and which APIs do not?

I think after that this would be good to merge

kiendang:
@lihaoyi readme and PR description updated.

kiendang:
Just found out you can make symlinks on windows. Wanna take a look at this first before finalizing.

kiendang:
@lihaoyi Need some help on deciding how zipping/unzipping symlinks would work on Windows here. You can create symlink on Windows, but only if (1) run as Administrator (by default) and (2) on NTFS.

Currently in this PR on Windows a symlink is
- always zipped as the referenced files (`followLinks = false` has no effect)
- unzipped as is, i.e. as a normal file with the name of the referenced file as its content, so a broken symlink.

This is similar to how Powershell's own [Compress-Archive](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/compress-archive?view=powershell-7.5) and [Expand-Archive](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/expand-archive?view=powershell-7.5) work.

Also tried Info-ZIP unzip, or more precisely the [Cosmo implementation](https://cosmo.zip/pub/cosmos/bin/) of it. There symlinks are
- zipped as symlinks or as the referenced files, i.e. following `followLinks`
- always unzipped as symlinks. If the symlink cannot be created an error is thrown and the file is not unzipped at all.

lihaoyi:
I think having `followLinks` sounds reasonable. We could fail with a nice error message if it is set to true on windows, and ask the user to set it to false?

kiendang:
Went on to implement `followLinks` for `unzip` and it turned out to be more complicated to do right than I thought. Simple way is to unzipping all the regular files and directories first, then for symlinks just copy the unzipped targets. However, this gets complicated when there are symlinks of symlinks, mutually referencing symlinks, self-referencing symlinks, ... We should be able to handle this with a loop/recursion, but would be more complex.

For now on Windows I just try unzipping symlinks as symlinks. If that doesn't work, unzip it as a file containing the target path (like what Expand-Archive did), and print out a message.

lihaoyi:
Sorry for not getting to this @kiendang, will take a look now

build.mill
```diff
@@ -194,6 +197,62 @@ object os extends Module {
...
+     * [[`jdk.zipfs` https://docs.oracle.com/en/java/javase/14/docs/api/jdk.zipfs/module-summary.html]] does not support
+     * symbolic links and only supports file permissions since JDK 14.
+     *
+     * Apache Ant `org.apache.tools.zip` was chosen over Apache Commons Compress due to the former not having any
+     * third party dependency, only depending on Java core libraries while the later also depends on Apache Commons IO.
+     *
+     * To avoid classpath conflicts, the dependency is shaded and compiled from source. Only the `org.apache.tools.zip`
+     * package, not the entire Ant codebase, is needed. This only adds < 100kb to Os-Lib jar size.
+     */
+    def apacheAntZipSource: T[PathRef] = Task {
```
lihaoyi, line 236: Can we make this a `def generatedSources` folder? That way we won't won't need to commit all the vendored code to the repo, and can rely on the build tool re-generating it on demand as necessary

os/src/ZipOps.scala
```diff
@@ -257,7 +325,53 @@ object unzip {
       excludePatterns: Seq[Regex] = List(),
       includePatterns: Seq[Regex] = List()
   ): os.Path = {
-    stream(os.read.stream(source), dest, excludePatterns, includePatterns)
+    checker.value.onWrite(dest)
+
```
lihaoyi, line 329: Could you remind me of the reason why we are unable to support permissions and symlinks in `unzip.stream`? I see that the upstream library has a version of `ZipOutputStream` in addition to `ZipFile`. Could we use that to get support for permissions and symlinks?

build.mill
kiendang: Yes

os/src/ZipOps.scala
kiendang: I've tried it and it doesn't work. It could be due to information on permissions and symlinks are stored as external attributes which are stored inside the zip central directory at the end of the zip file, not together with each zip entry. The Apache Commons Compress doc mentions [this](https://commons.apache.org/proper/commons-compress/zip.html) (in the ZipArchiveInputStream vs ZipFile section).

kiendang: > `ZipFile` is able to read the central directory first and provide correct and complete information on any ZIP archive.

lihaoyi: Got it, that's fine but we just ned to document this limitation somewhere. Maybe on the scaladoc for the `unzip.stream` class?

kiendang:
@lihaoyi Done.

I made an additional change to preserve permissions for zipped directories, not just files. When zipping, zip entries for directories (not just empty ones) are added. *e.g.* for `a/b/c/d.txt`, the following entries are added `a/b/`, `a/b/c/` and `a/b/c/d.txt`. This is similar to the Unix `zip` command.

I made shaded code `generatedSources`. Since it's all Java code I made the `os.zip` a `JavaModule`.

Adding a `followLinks` option to `os.unzip` to support unzipping symlinks as the referenced files is actually not as complicated as I previously thought. Do you still want it?

os/zip/src/os/shaded_org_apache_tools_zip/PermissionUtils.java
```diff
@@ -0,0 +1,153 @@
+/*
```
lihaoyi, line 1: @kiendang do we still need this one vendored file, or can it be removed with all the others?

kiendang: It comes from a different package, `org.apache.tools.ant.utils` instead of `org.apache.tools.zip`, but more importantly it requires modifying manually. There is some code that uses imports not from Java base libraries that needs to be removed. I did put a comment in the file.

lihaoyi: @kiendang got it!

lihaoyi:
I think this looks good, thanks @kiendang! I will transfer the bounty using your previous bank details

Issue #354, opened 2025-02-08
Expose a nice wrapper for `java.nio.file.Files.isX`


j-mie6:
The `Files.isReadable`, `Files.isWritable`, `Files.isExecutable`, etc methods would be nice to have exposed in os-lib. Something like `os.isReadable`, or `os.is.readable` would be nice.

j-mie6:
That said, these methods don't work well if the file doesn't exist. One might assume that something like this would work more predictably:

```scala
def writable(path: os.Path): Boolean = path match {
        case parent / _ if !os.exists(path) => writable(parent)
        case path => os.exists(path) && java.nio.file.Files.isWritable(path.toNIO)
    }
```

PR #376, opened 2025-03-10
Add isReadable, isWritable, isExecutable utilities

sake92:
Fixes https://github.com/com-lihaoyi/os-lib/issues/354


lihaoyi:
Thanks @sake92 !

PR #387, opened 2025-05-07
Make sure unzipped directories are owner-executable

kiendang:


lihaoyi:
Thanks @kiendang ! Seems like we may need to skip this on Windows which doesn't support POSIX permissions

kiendang:
I think we should add `OWNER_READ` as well?

lihaoyi:
@kiendang as long as we can unzip the problematic `classes.jar` we were hitting in Mill, any follow up permissions issues can be fixed at the callsite. Can you manually try unzipping that jar with your PR just to sanity check that this PR fixes the problem>

